PR #7, opened 2016-03-12
btrfs no space cache

abresas:
Sometimes after adding space with dd,  btrfs partition could not be expanded because of insufficient space.

A few users had hit this, and a user confirmed the problem went away after he rebooted, which makes me think it is some sort of caching issue.

This is an attempt to fix the issue by disabling free space cache on btrfs. As I have not been able to reproduce this, this is more like a guess at this point.


pcarranzav:
LGTM

petrosagg:
It'd be nice to have the PR description as the commit message. It will get lost otherwise

PR #13, opened 2016-05-10
sync and wait for mnt directory to exist before mounting to it

abresas:


preload.sh
```diff
@@ -85,6 +85,9 @@ losetup "$LOOP_DEV" "$IMAGE"
 partprobe "$LOOP_DEV"
 
 mkdir -p "/mnt/$APP_ID"
+# amke sure directory exists before mounting to it
```
Page-: `amke` -> `make`

preload.sh
```diff
@@ -85,6 +85,9 @@ losetup "$LOOP_DEV" "$IMAGE"
...
+sync
+sleep 2
```
Page-: I'm not a huge fan of the magic, but sgtm

PR #29, opened 2017-02-02
Fix calculation for expanding partition size

BrianAdams:
fix for https://github.com/resin-io/resin-preload-image-script/issues/28

### DIFF
--- a/src/preload.sh
+++ b/src/preload.sh
@@ -207,7 +207,7 @@ function get_resin_os_version() {
 # Usage: expand_image <image_path>
 function expand_image() {
     # Size will be increased by 110% of container size
-    IMAGE_ADD_SPACE=$(($CONTAINER_SIZE / 100 + 300))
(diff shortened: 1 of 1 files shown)


src/preload.sh
```diff
@@ -207,7 +207,7 @@ function get_resin_os_version() {
 # Usage: expand_image <image_path>
 function expand_image() {
     # Size will be increased by 110% of container size
-    IMAGE_ADD_SPACE=$(($CONTAINER_SIZE / 100 + 300))
+    IMAGE_ADD_SPACE=$(($CONTAINER_SIZE / 100 + $CONTAINER_SIZE))
```
Page-: 110% would be `$CONTAINER_SIZE / 10 + $CONTAINER_SIZE` or `$CONTAINER_SIZE * 1.1`

jhermsmeier: I'm going to merge this and change it to the intended 110% then

jhermsmeier:
Thanks @BrianAdams!

PR #40, opened 2017-05-01
preload.sh: Remove use of realpath

jhermsmeier:
The `realpath` is not available by default on several operating systems,
and hence this removes its usage to determine the dirname, as it should work without it as well.


jhermsmeier:
Closing in favor of #43

PR #47, opened 2017-05-30
Include app name in the generated apps.json to avoid an unnecessary restart

pcarranzav:
Change-Type: patch


pcarranzav:
Yeah, at some point the supervisor started also using the app name as something to compare to see if the app had to be updated... so not having a name makes the comparison fail and an "update" happen when the API responds with an app that has a name.

Issue #45, opened 2017-05-30
Default registry should be registry2.resin.io


pcarranzav:
Otherwise the device will be told to redownload from there causing an unwanted app restart

jhermsmeier:
Note to self: see [flowdock thread](https://www.flowdock.com/app/rulemotion/r-supervisor/threads/7nKF1gwXBJ8Vc9TLcy6OWwcUyW_)

jhermsmeier:
Fixed in #48 & #49

PR #48, opened 2017-06-09
src/preload.sh: Use registry2 in imageId

jhermsmeier:
This uses the registry2 host in the imageId written into the `apps.json`
preventing the supervisor downloading the app again on first boot.

Connects To: #45
Change-Type: patch

### DIFF
--- a/src/preload.sh
+++ b/src/preload.sh
@@ -271,7 +271,13 @@ function expand_btrfs() {
 # Write apps.json to file $1
 # Usage: write_apps_json <path>
 function write_apps_json() {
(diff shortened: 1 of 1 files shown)


pcarranzav:
The downloaded image also has to be tagged as `registry2.resin.io/.../...`

jhermsmeier:
@pcarranzav what exactly do you mean by that? I tested this with a RPi3 and it appears to work just fine.

pcarranzav:
Ah yeah, just realized this is done before pulling, so the image is properly tagged.

PR #49, opened 2017-06-12
src/preload.sh: Move to registry2

jhermsmeier:
This makes the preloader use the registry v2 to fix
the supervisor downloading preloaded applications again
on first boot.

Connects-To: #45, #48 
Change-Type: major

### DIFF
--- a/src/preload.sh
+++ b/src/preload.sh
@@ -5,7 +5,7 @@ set -o pipefail
 
 IMAGE=${IMAGE:-"/img/resin.img"}
 API_HOST=${API_HOST:-"https://api.resin.io"}
-REGISTRY_HOST=${REGISTRY_HOST:-"registry.resin.io"}
+REGISTRY_HOST=${REGISTRY_HOST:-"registry2.resin.io"}
 
 # Check if credentials have been set
 test "$API_TOKEN" -o "$API_KEY" || { echo >&2 "API_TOKEN or API_KEY must be set"; exit 1; }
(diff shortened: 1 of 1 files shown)


shaunmulligan:
@jhermsmeier and @pcarranzav I was testing this branch today, and it seems that the App is still restarted when the device finally connects to the internet. This is what I did:

* preload an `.dev` image (rpi3 one with no wifi credentials)
* boot device with serial cable attached, but not hooked up to ethernet yet.
* check that the user container is running correctly
* plug in ethernet cable, and see the user container gets restarted as soon as the device is registered and connected to the dashboard.

jhermsmeier:
I noticed the restart as well – @pcarranzav any idea what could be causing this? Is there something else the supervisor checks?

pcarranzav:
Hmm, very weird - what is the apps.json that you get? Does the supervisor also go through a "Downloading" stage or only kill and restart? @shaunmulligan @jhermsmeier

pcarranzav:
Ugh, I think it might be that the `RESIN_DEVICE_NAME_AT_INIT` variable doesn't really work as expected, and actually causes a restart when the name "changes" (i.e. when we connect to the API and actually get a value for the device name). I'll have to fix on the supervisor side.
(Though please still post an apps.json if you can, so that I can check it just in case)

jhermsmeier:
@pcarranzav only kill & restart – the downloading part is resolved by this (actually by #48 already).

I found this in supervisor: https://github.com/resin-io/resin-supervisor/blob/master/src/application.coffee#L262-L267 – could we be missing the `containerId`?

This is the `apps.json` this branch spits out:
```json
[{
  "appId": 211030,
  "name": "RPi3",
  "commit": "d4c62f0a3098db5a3fba66b631d3f32036617888",
  "imageRepo": "rpi3/d4c62f0a3098db5a3fba66b631d3f32036617888",
  "imageId": "registry2.resin.io/rpi3/d4c62f0a3098db5a3fba66b631d3f32036617888",
  "env": {},
  "config": {
    "RESIN_SUPERVISOR_DELTA": "1"
  }
}]
```

pcarranzav:
Nope, containerId isn't part of the comparison.
And the `RESIN_DEVICE_NAME_AT_INIT` theory I had doesn't seem to be either.
I'm trying to think what it could be but I'm a bit lost, I'll have to reproduce to see what may be wrong.
Do you happen to have supervisor logs of a device that hit this?

jhermsmeier:
@pcarranzav I'll spin up my RPi here again tomorrow with a freshly preloaded test image, and give you access so you can check it out, if you fancy?

PR #53, opened 2017-06-21
Support changing the boot splash image.

cacsar:
Add commandline option --splash-image to support changing the boot
splash image.

### DIFF
--- a/preload.sh
+++ b/preload.sh
@@ -22,6 +22,7 @@ function usage() {
     echo "  --api-key      API key"
     echo "  --api-host     API hostname"
     echo "  --registry     Image registry host"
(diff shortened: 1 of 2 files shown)


jhermsmeier:
Would you like to also update the `README.md`, adding the option? I can do that later as well, though.

cacsar:
I took a stab at updating the README to do it, but I wasn't able to get it written to my reasonable satisfaction (so I'll leave it to you if that's alright). Good documentation would reference the docs referenced by https://github.com/resin-io/docs/issues/155 . https://github.com/resin-io/resin-plugin-img is probably another route of doing the functionality I added here.

jhermsmeier:
No worries, I'll do that then :)

Issue #54, opened 2017-06-23
Running out of space on image during preload


BrianAdams:
When I download the docker image to my desktop to check the size I see that is it *1.91GB*, yet when doing the resin-preload, the script calculates the size as *798MB*.  I am assuming that is why it is running out of space when downloading the image:

```
docker images
REPOSITORY                                                                     TAG                 IMAGE ID            CREATED             SIZE
registry2.resin.io/...project.../86a155bc9b33e6f1d9710063638e26dc22921f42   latest              0cc797f6ebff        2 days ago          1.91GB
```

```
resin os download raspberrypi3 -o resin.img --version default
docker run --rm -e API_KEY=..key.. -e APP_ID=..appid.. -v /tmp/resin.img:/img/resin.img --privileged resin/resin-preload
```

```
Fetching application data
Using API host https://api.resin.io
Using Registry host registry2.resin.io
Using API_KEY
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   840  100   840    0     0    874      0 --:--:-- --:--:-- --:--:--  1746
...
Container size: 798 MB
Mapping /img/resin.img 46137344:327155712 to /dev/loop0

Mounting rootfs from /dev/loop0 to /mnt/202699-rootfs

Sourcing release info from /mnt/202699-rootfs/etc/os-release
Detected Resin OS 2.0.6+rev3

id: resin-os
name: Resin OS
version: 2.0.6+rev3
version_id: 2.0.6
pretty_name: Resin OS 2.0.6+rev3
resin_board_rev: 5f3fbb3
meta_resin_rev: c653cc2
slug: raspberrypi3
machine: raspberrypi3


Unmounting rootfs from /mnt/202699-rootfs
umount: /mnt/202699-rootfs unmounted
Unmapping /dev/loop0
Model:  (file)
Disk /img/resin.img: 1804MB
Sector size (logical/physical): 512B/512B
Partition Table: msdos
Disk Flags:

Number  Start   End     Size    Type      File system  Flags
 1      4.19MB  46.1MB  41.9MB  primary   fat16        boot, lba
 2      46.1MB  373MB   327MB   primary   ext4
 3      373MB   700MB   327MB   primary
 4      700MB   1804MB  1103MB  extended               lba
 5      705MB   726MB   21.0MB  logical   ext4
 6      730MB   1804MB  1074MB  logical   ext4

[...]

PR #55, opened 2017-06-23
src/preload.sh: Force-disable compression for container blob

jhermsmeier:
This sets `Accept-Encoding: identity` to avoid having compressed `Content-Length` values for blob headers, causing inaccurate container size calculation.

Connects To: #54


jhermsmeier:
As it turns out, forcing `identity` content-encoding does nothing for the Docker Registry, as the identity is still a `.tar.gz` – closing.

PR #56, opened 2017-06-28
src/preload.sh: Fix condition checking for /rce dir

jhermsmeier:
This fixes a condition checking for a legacy rce setup,
which would've caused those to not be properly cleaned up.

Change-Type: patch

### DIFF
--- a/src/preload.sh
+++ b/src/preload.sh
@@ -314,8 +314,10 @@ function start_docker_daemon() {
     # If this preload script was ran before implementing the rce/docker fix,
     # make sure you cleanup
     if [[ -d "${APPFS_MNT}/docker" ]]; then
-        log "Removing" "${APPFS_MNT}/rce"
-        rm -rfv "${APPFS_MNT}/rce"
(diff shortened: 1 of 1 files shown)


src/preload.sh
```diff
@@ -313,7 +313,7 @@ function start_docker_daemon() {
 
     # If this preload script was ran before implementing the rce/docker fix,
     # make sure you cleanup
-    if [[ -d "${APPFS_MNT}/docker" ]]; then
+    if [[ -d "${APPFS_MNT}/rce" ]]; then
```
Page-: I think the previous version is correct, what this does is check if it's an image using `docker` and then removes any stray `rce` stuff that may have been added if running with an old version of the preloader that assumed `rce` only, and then sets the docker dir to `"${APPFS_MNT}/docker"`

jhermsmeier: Oh, right – got confused by the output there. I've now reverted that, and added another condition within this, so that it only prints & removes the `/rce` directory if it exists.

PR #60, opened 2017-07-04
src/preload.sh: Double-quote vars to prevent globbing / word splitting

jhermsmeier:
This mends some `shellcheck` warnings about double-quoting vars to prevent globbing & word-splitting.

Change-Type: patch

### DIFF
--- a/src/preload.sh
+++ b/src/preload.sh
@@ -35,42 +35,42 @@ DOCKER_DIR=
 # Log to stderr to avoid interfering
 # with function output
 function log() {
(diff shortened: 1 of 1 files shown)


src/preload.sh
```diff
@@ -44,33 +44,33 @@ function cleanup() {
 
     if [[ -e $DOCKER_PID ]]; then
```
Page-: This should also have quotes

jhermsmeier: Fixed!

Issue #51, opened 2017-06-14
Make sure image path leads to a file before preloading


jhermsmeier:
If the image path given does not resolve to a file, the mapping to the docker container fails, but doesn't error, and causes a fail of `losetup` later in the process.

To avoid this, the image path should be `stat`ed early, and the preloader should error immediately if it does not resolve to a file.

zvin:
fixed in #62

PR #62, opened 2017-07-07
Make sure --img argument is a file accessible for reading and writing before preloading.

zvin:
Change-Type: patch

### DIFF
--- a/bin/resin-preload
+++ b/bin/resin-preload
@@ -114,6 +114,24 @@ if( !(options.appId && options.image && (options.apiToken || options.apiKey)) ) 
   process.exit(1)
 }
 
+const fs = require('fs')
+
+function isReadableAndWritable(path) {
+  try {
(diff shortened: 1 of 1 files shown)


zvin:
Connects to #51

bin/resin-preload
```diff
@@ -114,6 +114,24 @@ if( !(options.appId && options.image && (options.apiToken || options.apiKey)) )
   process.exit(1)
 }
 
+const fs = require('fs')
+
+function isReadableAndWritable(path) {
+  try {
+    fs.accessSync(options.image, fs.constants.R_OK | fs.constants.W_OK)
```
jhermsmeier: Looks like `options.image` should be `path` here

zvin: good catch, fixed

jhermsmeier:
@zvin don't forget to bump the version & publish ;)

zvin:
@jhermsmeier done: https://www.npmjs.com/package/resin-preload

PR #65, opened 2017-07-19
switch preload script from bash to python3

zvin:
This makes it readable and paves the way for supporting flasher type images.


src/preload.py
```diff
@@ -0,0 +1,329 @@
...
+
+def get_registry_token(image_repo):
+    params = {
+        "service": REGISTRY_HOST,
+        "scope": "repository:{}:pull".format(image_repo),
+    }
+    return api("auth/v1/token", params)["token"]
+
+
+def registry(endpoint, registry_token, headers=None, json=True):
```
imrehg: Probably rename this `json` to something else, since it redefining name 'json' from outer scope

Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM docker:1.10-dind
+FROM docker:17.06.1-ce-rc1-dind
 
-RUN apk update && apk add --no-cache bash curl jq parted btrfs-progs docker util-linux
+RUN apk update && apk add --no-cache python3 parted btrfs-progs docker util-linux inotify-tools
+RUN pip3 install sh==1.12.14 requests==2.18.1
```
imrehg: Can the python dependencies be put in a `requirements.txt` and installed from there? It would be more maintainable/transparent, than having it inline installed in the `Dockerfile`, imho.

src/preload.py
```diff
@@ -0,0 +1,329 @@
...
+    suffixes = ['', 'Ki', 'Mi', 'Gi', 'Ti']
+    idx = 0
+    while idx < 4 and size >= 1024:
+        idx += 1
+        size /= 1024
+    result = "{:.{}f}".format(size, precision).rstrip("0").rstrip(".")
+    return "{} {}B".format(result, suffixes[idx])
+
+
+def log(*args):
```
imrehg: Question, could standard `logging` be used for logs, to achieve the same effect? I feel it would be nicer (even though this current one is lightweight)

src/preload.py
```diff
@@ -0,0 +1,329 @@
...
+    # Add zero bytes to image to be able to resize partitions
+    with open(image, "a") as f:
+        size = f.tell()
+        f.truncate(size + additional_space)
+
+
+def expand_partitions(image):
+    # Resizes partition 4 & 6 to the end of the image
+    log("Expanding extended partition 4 to the end of the disk image.")
+    log("Expanding logical partition 6 to the end of the disk image.")
```
imrehg: Since below it's a single operation, it would be good to have a single line of log as well (multiple log lines to me mean multiple tasks, and things happening between log lines)

src/preload.py
```diff
@@ -0,0 +1,329 @@
...
+    headers = headers or {}
+    headers["Authorization"] = "Bearer {}".format(registry_token)
+    url = "https://{}/{}".format(REGISTRY_HOST, endpoint)
+    response = get(url, headers=headers)
+    response.raise_for_status()
+    return response.json() if json else response.content
+
+
+def get_app_data():
+    # Fetch application metadata
```
imrehg: Instead of comments like this, can proper docstrings be added, and done for each function?

src/preload.py
zvin: updated

Dockerfile
zvin: requirements moved to requirements.txt

src/preload.py
zvin: done, it's a bit more verbose

src/preload.py
zvin: updated

src/preload.py
zvin: existing comments converted

src/preload.py
```diff
@@ -0,0 +1,342 @@
...
+            else:
+                log.warning("e2fsck: File system errors corrected")
+        except ErrorReturnCode:
+            log.error("e2fsck: File system errors could not be corrected")
+            exit(1)
+        resize2fs("-f", loop_device)
+
+
+def expand_btrfs(mountpoint):
+    # For btrfs we'll need to mount the fs.
```
jhermsmeier: This comment seems to be in the wrong place, as the line below it isn't doing what the comment says ;)

src/preload.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def get_app_data():
+    """Fetches application metadata"""
+    endpoint = "v2/application({})?$expand=environment_variable".format(APP_ID)
+    data = api(endpoint)["d"][0]
+    image_repo = "{app_name}/{commit}".format(**data).lower()
+    image_id = "{}/{}".format(REGISTRY_HOST, image_repo).lower()
+    env = {v["name"]: v["value"] for v in data.get("environment_variable", [])}
+    return {
```
shaunmulligan: I think this needs to return an array as per the flowdock discussion here: https://www.flowdock.com/app/rulemotion/r-supervisor/threads/bTG6bDJF-dCiIo7B7WG_6n4_9Yz

zvin: @shaunmulligan fixed in https://github.com/resin-io/resin-preload/pull/65/commits/165f22ed53ef4a41008c1c37de32aee570a18c85

src/preload.py
zvin: @jhermsmeier comment removed, there is a similar comment here https://github.com/resin-io/resin-preload/blob/bba6110f92e65c2424c5d8aabff38b3b2b1dc7ae/src/preload.py#L301

zvin:
@jhermsmeier Python seemed a good fit here, especially with https://pypi.python.org/pypi/sh , I just wanted to remove the need for `jq` and `curl` and global mutable variables.
This could have been done with Node but IMO it would have been slightly less readable.

Issue #37, opened 2017-03-30
Support flasher type images


jhermsmeier:
Support "flasher type" images which flash to internal storage devices (eMMC, etc.).
Note: Images of this type are used for Beaglebones, NUCs and UP boards.

- [x] Detect whether it's a "flasher type" image
- [x] Read `yocto.deployArtifact` key from `device-type.json`, replace `'-flasher'' to get filename
- [x] move the partitions 3, 4, 5 and 6 to make space for partition 3
- [x] resize the partition 2
- [x] mount the `${filename}.resinos-img` from `/opt` of the 2nd partition (rootfs)
- [x] resize it to fit the application container
- [x] preload application into that

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_1i41)

floion:
Hi, any progress on this?

zvin:
for point 2: `jq -r '.yocto.deployArtifact' boot/device-type.json | sed s/flasher-//`

zvin:
@floion I'm working on it, but I'll be on vacation until July 18th.

PR #66, opened 2017-07-20
Support flasher type images

zvin:
This requires #65 to be merged first.
Connects to #37 

How it works:

 * look at the `deployArtifact` key in the `device-type.json` on the 1st partition of the image. If it contains "-flasher-" continue to the next steps, otherwise preload to the image as before.
 * resize the 2nd partition of the outer image (and move the partitions after) so the inner image fits
 * resize and preload the inner image (/opt/<deployArtifact stripped of "flasher-"> in the second partition of the outer image)

For resizing the 2nd partition and moving the ones after we use `sfdisk` to generate an sfdisk script of the image layout, update it and create a new image with the updated layout. Then we dd partitions from the original image to the new one, resize the ext filesystem of the 2nd partition and replace the original image with the one we just created.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM docker:17.06.1-ce-rc1-dind
+FROM docker:17.06.1-ce-dind
 
-RUN apk update && apk add --no-cache python3 parted btrfs-progs docker util-linux inotify-tools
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,9 +1,13 @@
-FROM docker:1.10-dind
+FROM docker:17.06.1-ce-rc1-dind
```
zvin: None, I'll update the Dockerfile.

PR #70, opened 2017-07-27
Create an empty CHANGELOG.

zvin:
Change-Type: patch

Connects to #63
---- Autogenerated Waffleboard Connection: Connects to #63


zvin:
I had to create a new version because the tag v1.2.0 already exists.

package.json
```diff
@@ -33,4 +33,4 @@
   "bin": {
     "resin-preload": "bin/resin-preload"
   }
-}
+}
```
jhermsmeier: Looks like there's a missing EOL here?

zvin: Looks like yes, the branch is already merged, I'll fix in in the next PR.

PR #73, opened 2017-07-28
Use standardJS

zvin:
Connects to #72

Change-Type: patch


.editorconfig
```diff
@@ -7,11 +7,11 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
-indent_style = space
-indent_size = 2
+indent_style = tab
+indent_size = 4
```
jhermsmeier: Why this change? For JS most of our repos use the 2 space indent, I think it'd be better to stick with that.

zvin: I always thought that we preferred tabs instead of spaces in resin. I'll set the size back to 2 but can we keep the tabs ?

zvin: After checking some resin repos, you are right, most of them use spaces for js. I'll change that back.

zvin: done

bin/resin-preload
```diff
@@ -97,57 +101,53 @@ const options = {
...
-    case '--registry': options.registryHost = argv.shift(); break
-    case '--splash-image': options.splashImage = argv.shift(); break
+  case '--app': options.appId = argv.shift(); break;
+  case '--img': options.image = argv.shift(); break;
+  case '--api-token': options.apiToken = argv.shift(); break;
+  case '--api-key': options.apiKey = argv.shift(); break;
+  case '--commit': options.commit = argv.shift(); break;
+  case '--api-host': options.apiHost = argv.shift(); break;
+  case '--registry': options.registryHost = argv.shift(); break;
+  case '--splash-image': options.splashImage = argv.shift(); break;
```
jhermsmeier: There's one level of indentation missing here

zvin: That was the default rules of eslint, I changed it now. (I also added the missing .eslintrc file).

bin/resin-preload
```diff
@@ -34,58 +39,57 @@ const USAGE = `
...
-    error.signal = signal
-    showError( error )
-    process.exit( code || 1 )
+    const error = new Error(
+      `Preloader exited with ${code == null ? 'signal' : 'code'} ${code == null ? signal : code}`
+    );
+    error.code = code;
+    error.signal = signal;
+    showError( error );
+    process.exit( code || 1 );
```
jhermsmeier: This still seems inconsistent with the style elsewhere (spaces inside parens)

zvin: Let's wait until the precess call tomorrow.

zvin:
@jviotti @jhermsmeier @lekkas 

I tried to use Etcher's .eslintrc file.
I find it way to strict:
 * it forces me to install lodash even if the project wasn't using it before because it prefers `_.assign` to `Object.assign`;
 * the no-magic numbers rule is pretty annoying, for example this is considered an error:
```js
if (compareVersions(dockerApiVersion, '1.28') >= 0) {
```
 * no-underscore-dangle combined with no-param-reassign is annoying;
 * no-bitwise makes no sense at all; why would we forbid that ? useful example:
```js
fs.accessSync(path, fs.constants.R_OK | fs.constants.W_OK);
```

I can adapt the .eslintrc file to my needs but the point was to use something standard between projects.

jhermsmeier:
@zvin if I understood it correctly yesterday we agreed on just using standardJS as the base config, so no need to pull in Etcher's eslint config anymore. Etcher itself now uses standard, only extending it with its stricter rules.

zvin:
@jhermsmeier thanks for the info, I'm switching this to standardJS.

package.json
```diff
@@ -28,9 +30,10 @@
     "url": "https://github.com/resin-io/resin-preload/issues"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "pretest": "standard lib/preload.js bin/resin-preload",
+    "test": "echo done"
```
jhermsmeier: You can omit the `test` script here, and it will just echo 'Error: no test specified' but still exit with a zero status code

zvin: done

.circleci/config.yml
```diff
@@ -0,0 +1 @@
+version: 2
```
jhermsmeier: Does CircleCI not need more config than this? What about testing multiple Node versions?

jhermsmeier: And why are we switching to CircleCI in this PR (and at all)? This should happen in a separate one, as this one should only be concerned with eslint.

zvin: We're not switching.
We had nothing before.
I'm adding this following your comment https://github.com/resin-io/resin-preload/pull/73#pullrequestreview-53270923 .

zvin: I'm not sure it requires more settings, we'll see once it is enabled.
This is just for running `standard` on the 2 js files we have.
There are no tests so testing multiple node versions makes no sense.

jhermsmeier: Pretty much every resin project runs on TravisCI (apart from something recently switching to CircleCI), let's stick with Travis for the time being, as access to it etc. is already set up & sorted for the entire team. And then we can look at switching CI services separately.

zvin: @jhermsmeier please see the conversation here: https://www.flowdock.com/app/rulemotion/r-operations/threads/5GW1zzD0MXZRYo22OPdW7PlOtes

CircleCI is the way to go now.
You don't need to set up anything specific in CircleCi, just log in with your github account.

zvin: @jhermsmeier https://circleci.com/gh/resin-io/resin-preload/4

jhermsmeier: Okay, don't mind me then – sorry for the confusion

hedss:
Sorry, I've probably missed this in the Architecture call. The last thing discussed in the process call was using ESlint, so was it agreed in the arch call to move to standard?

zvin:
@hedss I wasn't at this call either and I can't find the notes but that's what @jhermsmeier said.

PR #81, opened 2017-08-04
Integrate with resin-cli

zvin:
Almost ready but some things are still missing.

 * use dockerode instead of starting docke sub-processes;
 * interactive menu for selecting the app and commit to preload;
 * ask user and disable automatic app updates if he chooses to preload a commit that is not the latest;
 * detect device-type from the image and allow preloading only apps that have the same device type;
 * pass the RESINRC_PROXY env var to the preload container as HTTP(S)_PROXY
 * fix typo

Connects to #71

Change-Type: minor


lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+
+function disableAutomaticUpdates (form, resin, application, commit) {
+  if ((commit === LATEST) || !application.should_track_latest_release) {
+    return Promise.resolve()
+  }
+  const message = (
+    'This application is set to automatically update all devices to the ' +
+    'latest available version.\n' +
+    'You need to disable this if you want to preload a specific build. ' +
+    'Do you want to disable automatic updates for this application?'
```
CameronDiver: This line would be 128 characters when printed to the CLI, it would be better to shorten it.

lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+    choices: [
+      {name: 'Yes', value: true},
+      {name: 'No', value: false}
+    ]
+  })
+  .then(function (update) {
+    if (!update) {
+      return
+    }
+    return resin.pine.patch({
```
CameronDiver: It would be nice if these functions could be added to the SDK. If they are already it would be nice to use them.

lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+  description: '(beta) preload an app on a disk image',
+  help: `Warning: 'resin preload' requires Docker to be correctly installed in
+your shell environment. For more information (including Windows support)
+please check the README here: https://github.com/resin-io/resin-cli .
+
+Use this command to preload an application to a local disk image.
+
+Examples:
+  $ resin preload resin.img --app 12345 --commit e1f2592fc6ee949e68756d4f4a48e49bff8d72a0 --splash-image some-image.png
+  $ resin preload resin.img`,
```
CameronDiver: This will present `resin preload` as the interface when not in the cli, right?

lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+      alias: 'c'
+    },
+    {
+      signature: 'splash-image',
+      parameter: 'splashImage.png',
+      description: 'path to a png image to replace the splash screen',
+      alias: 's'
+    },
+    {
+      signature: 'dont-detect-flasher-type-images',
```
CameronDiver: This option feels a bit verbose, I think at the very least we should add an alias for it,

lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+      boolean: true,
+      description: 'Disables the flasher type images detection: treats all images as non flasher types'
+    }
+  ],
+  action: function (params, options, done) {
+    console.log(options)
```
CameronDiver: Debugging code, don't forget to remove.

lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+
+    options.image = params.image
+    options.appId = options.app
+    delete options.app
+
+    options.dontDetectFlasherTypeImages = options['dont-detect-flasher-type-images']
+    delete options['dont-detect-flasher-type-images']
+
+    return preload.build()
+    .then(resin.settings.getAll)
```
CameronDiver: The promise chain from this point onwards should be indented, because of the `return`.

lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+    .then((settings) => {
+      console.log(settings)
```
CameronDiver: Debugging, don't forget to remove.

lib/capitano.js
```diff
@@ -0,0 +1,188 @@
...
+        }
+        return selectApplication(resin, form, deviceType)
+      })
+      .then((application) => {
+        options.appId = application.id
+        if (deviceType !== application.device_type) {
+          throw new Error(`Image device type (${application.device_type}) and application device type (${deviceType}) do not match`)
+        }
+        return resin.models.build.getAllByApplication(application.id)
+        .then((builds) => {
```
CameronDiver: Wonky indentation here.

lib/preload.js
```diff
@@ -1,32 +1,125 @@
...
-const escape = require('command-join')
-const compareVersions = require('compare-versions')
+const Docker = require('dockerode')
+const tarfs = require('tar-fs')
+const stream = require('stream')
+const Promise = require('bluebird')
 
 const preload = module.exports
 
+const docker = new Docker({Promise: Promise})
```
CameronDiver: This can be 
```
const docker = new Docker({Promise})

File: lib/preload.js [hunk 10]
```diff
@@ -1,32 +1,125 @@
 'use strict'
 
-const childProcess = require('child_process')
 const path = require('path')
-const escape = require('command-join')
-const compareVersions = require('compare-versions')
+const Docker = require('dockerode')
+const tarfs = require('tar-fs')
```
CameronDiver: I would watch out for using this module. The underlying `tar-stream` module has a bug when using the streaming interface which can cause hanging. To solve this it's best to use `tar-stream` directly, with the `Buffer` interface (but it is less efficient).

lib/preload.js
```diff
@@ -1,32 +1,125 @@
...
-  }, processOptions)
-
-  const dirname = path.join(__dirname, '..')
-  const argv = [ 'build', '-t', 'resin/resin-preload', dirname ]
-
-  return childProcess.spawn('docker', argv, processOptions)
+preload.build = function () {
+  console.log('Building preloader image...')
+  return docker.buildImage(
+    tarfs.pack(path.resolve(__dirname, '..')),
```
CameronDiver: There is functions in resin-cli which already do this, which are also hardened against the bug I mentioned.

lib/preload.js
```diff
@@ -1,32 +1,125 @@
...
+            return
+          }
+          resolve()
+        },
+        (event) => {
+          if (event.stream) {
+            process.stdout.write(event.stream)
+          }
+        }
+      )
```
CameronDiver: This function is pretty hard to parse, maybe a couple of comments or different layout would help?

lib/preload.js
```diff
@@ -1,32 +1,125 @@
...
+const bindMount = (source, target) => {
+  return {
+    Source: path.resolve(source),
+    Target: target,
+    Type: 'bind',
+    Consistency: 'delegated'
+  }
+}
+
+const createContainerDisposer = (options) => {
```
CameronDiver: Really nice idea :+1: I feel like the builder could benefit from this kind of interface!

lib/preload.js
```diff
@@ -1,32 +1,125 @@
...
+    return container.remove()
+  })
+}
+
+const runContainer = (container) => {
+  return container.start()
+  .then((container) => {
+    return container.attach({stream: true, stdout: true, stderr: true})
+  })
+  .then((stream_) => {
```
CameronDiver: Why are you using `stream_` for a variable name? It seems like if you're avoiding name clashes, it would be better to use something like `dockerStream` or `buildStream`.

lib/capitano.js
zvin: This may be integrated to the SDK some day: https://www.flowdock.com/app/rulemotion/resin-tech/threads/wI7v5qGGkpzER6MvaVHkVY-HP6k

lib/capitano.js
zvin: I'm not sure I get the question.
This will be the interface when using resin-cli.
The standalone interface remains bin/resin-preload.

lib/capitano.js
zvin: This is a very specific option to use only when you know what you're doing. It only has one user for now. I'd rather not add it an easy alias.

lib/capitano.js
CameronDiver: You could probably add it yourself if you're feeling generous :p

lib/capitano.js
CameronDiver: So the standalone interface doesn't even use capitano? This is definitely something we should look at.

lib/capitano.js
zvin: I won't :)

lib/capitano.js
CameronDiver: Fair enough, I don't feel that strongly about it, it just feels unnecessarily verbose even if people shouldn't be using it.

lib/capitano.js
zvin: I didn't use that.
My convention is that only a `(`, a `{` or a `[` allow me to indent the next line.
I find it a bit better looking.
If we want to change it to indent the `then`s, it should be done everywhere and probably enforced by the linter.

lib/preload.js
zvin: That was because of the imported `stream` module.
I renamed it to `streamModule`.

lib/capitano.js
CameronDiver: I was pretty sure it was in the standard as I've been called up on it a few times by @Page- . It's used in quite a few backend components so I think we should be adding it. Also I think it does increase readability a lot.

It should be our convention, not my convention :)

zvin: Example:

I find that this

```javascript
preload.build = function () {
  console.log('Building preloader image...')
  return docker.buildImage(
    tarfs.pack(path.resolve(__dirname, '..')),
    {t: DOCKER_IMAGE_TAG}
  )
  .then((stream) => {
    return new Promise((resolve, reject) => {
      docker.modem.followProgress(
        stream,
        (err, output) => {
          if (err) {
            reject(err)
            return
          }
          resolve()
        },
        (event) => {
          if (event.stream) {
            process.stdout.write(event.stream)
          }
        }
      )
    })
  })
}
```

looks better than

```javascript
preload.build = function () {
  console.log('Building preloader image...')
  return docker.buildImage(
    tarfs.pack(path.resolve(__dirname, '..')),
    {t: DOCKER_IMAGE_TAG}
  )
    .then((stream) => {
      return new Promise((resolve, reject) => {
        docker.modem.followProgress(
          stream,
          (err, output) => {
            if (err) {
              reject(err)
              return
            }
            resolve()
          },
          (event) => {
            if (event.stream) {
              process.stdout.write(event.stream)
            }
          }
        )
      })
    })
}
```

CameronDiver: I feel like this is quite a contrived example as that is a block of spaghetti code :P (JS's fault, not yours)

lib/preload.js
zvin: Added a few comments, docker.modem.followProgress takes 2 callbacks: `onProgress` and `onFinished`. It should be a bit more clear now.

zvin: I could use named functions if you prefer.

lib/capitano.js
jhermsmeier: What is this doing? Is `event.stream` actually a stream, or a string?

```js
if (event.stream) {
  process.stdout.write(event.stream)
}
```

zvin: That's a string, it's the docker output.

zvin: like `Step 1/7 : FROM docker:17.06.1-ce-dind`

zvin: `docker.modem.followProgress` docs: https://github.com/apocas/dockerode#helper-functions

jhermsmeier: Apart from that, I'd highly recommend not wrapping streams in Promises; I'd argue the following is much simpler to read & deal with:

```js
preload.build = function () {
  
  var stream = new stream.PassThrough()
  var pack = tarfs.pack(path.resolve(__dirname, '..'))
  
  docker.buildImage( pack, { t: DOCKER_IMAGE_TAG })
    .catch((error) => { stream.emit('error', error) })
    .then((build) => {
      docker.modem.followProgress(build, (error, output) => {
        if (error) {
          stream.emit('error', error)
        }
      }, (event) => {
        stream.write(event.stream)
      })
    })
  
  return stream
  
}
```

zvin: @jhermsmeier I have to disagree here, the caller doesn't need a stream, it just needs a Promise of when the build will be done.

zvin: @CameronDiver more on why I don't like indenting `.then`s:
 * we don't do it in coffeescript, why should we in js ?
 * indenting on the next line after returns creates some missing closing brackets / parenthesis: see the attached image, to me it seems like something is missing near the red dot.
capture d ecran de 2017-08-04 19-48-49

zvin: Reading only the bottom part you could think that some closing bracket / parenthesis are missing.

lib/preload.js
zvin: The problem with importing from resin-cli in lib/preload.js is that it forces us to either
 * remove the standalone bin/resin-preload
 * make resin-preload depend on resin-cli

Maybe we could move this function to [docker-toolbel](thttps://github.com/resin-io-modules/docker-toolbelt/)?

lib/capitano.js
jhermsmeier: > I have to disagree here, the caller doesn't need a stream, it just needs a Promise of when the build will be done.

This module is installable by other consumers than our CLI, and the caller should receive a stream of the child process' data, instead of it writing out to stdout, so it can handle & process the data.

zvin: @jhermsmeier that's a valid point but it leaves me with some questions:
 * the point of resin-preload os to preload, not to show how it does it, should we really be returning a stream of docker output?
 * if we do this with build, I guess we should do the same with run?
 * what about the things like `console.log('Building preloader image...')` ? Should we keep them, remove them, or include them in the returned stream?

zvin: Answer to myself for the 3rd point: move them to capitano.js and bin/resin-preload.

zvin: @jhermsmeier I guess I have no more objections, I'll change `build` to return a stream and `run` to return `[stdout, stderr]`.
I can always have separate helpers that convert these streams to promises.

jhermsmeier: Okay, cool – as this changes the API of the exported methods, the Change-Type will need a bump to "major" I just realized

CameronDiver: @zvin We do tend to do it in CS, but it's not in our coding standards for whatever reason so I'm not going to block you on it. I would ask though if you make changes to a codebase which does follow this convention, that you also follow this convention.

lib/preload.js
CameronDiver: @zvin that sounds fine to me, if docker-toolbelt no longer does dirty mutation :p

lib/preload.js
CameronDiver: Ah I didn't see that, fair enough, SGTM.

zvin:
@CameronDiver @emirotin I think I'll move lib/capitano.js to `resin-cli`. It has no other use and I'd like to use some utility functions that are in `resin-cli`.
Any objection?

CameronDiver:
@zvin I'm all for that :+1:

emirotin:
Yeah I was going to ask if that's what should be done as it belongs to the CLI integration

emirotin:
> pass the RESINRC_PROXY env var to the preload container as HTTP(S)_PROXY

You shouldn't rely on this specific env var (RESINRC_PROXY) as there are multiple ways to set the ptoxy. Instead use the resin settings module, and check the code in the CLI responsible for detecting the proxy config

zvin:
@emirotin I used settings.proxy: https://github.com/resin-io/resin-cli/blob/8684b65a42a4fd6d1d3c58cf4b1fdf26b959b682/lib/actions/preload.coffee#L171

bin/resin-preload
```diff
@@ -2,9 +2,12 @@
 
 'use strict'
 
-const fs = require('fs')
+const Docker = require('dockerode')
 const preload = require('..')
 const info = require('../package.json')
+const streamToPromise = require('stream-to-promise')
+
+const docker = new Docker({Promise})
```
CameronDiver: This will only ever use the local docker, which I assume is fine for this use-case, just thought i'd let you know.

That is also assuming that the default is `{ socketPath: '/var/run/docker.sock' }` and whatever the equivalent on windows is, it may be worth adding this in anyway.

bin/resin-preload
```diff
@@ -125,30 +118,21 @@ if (!(options.appId && options.image && (options.apiToken || options.apiKey))) {
...
-  preload.run(options, { stdio: 'inherit' }).once('exit', (code, signal) => {
-    exitIfError(code, signal)
-    console.log('\n[OK] Done.')
-  })
+  return preload.run(docker, options)
+})
+.then((streams) => {
+  streams[0].pipe(process.stdout)
+  streams[1].pipe(process.stderr)
+  return Promise.all(streams.map(streamToPromise))
```
CameronDiver: Can you use `Promise.map` here?

lib/preload.js
```diff
@@ -1,32 +1,182 @@
...
+const createContainerDisposer = (docker, options) => {
+  return createContainer(docker, options)
+  .disposer((container) => {
+    return container.remove()
+  })
+}
+
+const runContainer = (docker, container) => {
+  const killContainer = (data) => {
+    if (data[0] !== 3) {
```
CameronDiver: It may be worth exporting this magic number to a const value.

lib/preload.js
```diff
@@ -1,32 +1,182 @@
...
+}
+
+const createContainerDisposer = (docker, options) => {
+  return createContainer(docker, options)
+  .disposer((container) => {
+    return container.remove()
+  })
+}
+
+const runContainer = (docker, container) => {
```
CameronDiver: This whole function needs comments, why are we setting raw mode for example?

lib/preload.js
```diff
@@ -1,32 +1,182 @@
...
+  })
+  .then((stream) => {
+    docker.modem.demuxStream(stream, stdout, stderr)
+    return [stdout, stderr]
+  })
+}
+
+const isReadableAndWritable = (image) => {
+  try {
+    fs.accessSync(image, fs.constants.R_OK | fs.constants.W_OK)
```
CameronDiver: Why are you using synchronous access here?

lib/preload.js
```diff
@@ -1,32 +1,182 @@
...
+  } catch (e) {
+    return false
+  }
+  return true
+}
+
+const checkImage = (image) => {
+  const rwAccess = isReadableAndWritable(image)
+  if (!(rwAccess && fs.statSync(image).isFile())) {
```
CameronDiver: Synchronous file accesses, these should all be async unless there is a good reason not to (and I don't see any comments saying why it is necessary).

bin/resin-preload
zvin: That's only for the standalone bin/resin-preload . I didn't change it.
The new resin-cli preload way allows you to specify a docker host or socket.

bin/resin-preload
zvin: done

bin/resin-preload
CameronDiver: Yeah I figured as much, and guessed dockerode would default to the local docker socket.

lib/preload.js
zvin: It was to keep the code simple. I changed it to use Async methods.

lib/preload.js
zvin: Same as above, changed to Async methods now.

lib/preload.js
zvin: comments added, raw mode was tor be able to intercept ctrl-c

lib/preload.js
zvin: done, that was CTRL_C

lib/preload.js
```diff
@@ -1,32 +1,196 @@
...
+    const statusCodePromise = new Promise((resolve, reject) => {
+      // Delete the container once it ends (or gets killed).
+      container.wait()
+      .then((info) => {
+        removeListener()
+        container.remove()
+        resolve(info.StatusCode)
+      })
+      .catch(reject)
+    })
```
CameronDiver: I'm pretty sure you can make this:
```javascript
const statusCodePromise = container.wait()
.then((info) => {
  removeListener()
  container.remove()
  return info.StatusCode
})
```

Also do we need to be waiting on `container.remove`?

zvin: Good point, updated.

PR #85, opened 2017-08-16
Correctly set the resin api token.

zvin:
Connects to #84

Change-Type: patch


bin/resin-preload
```diff
@@ -132,6 +131,12 @@ buildOutputStream.pipe(process.stdout)
 
 streamToPromise(buildOutputStream)
 .then(() => {
+  if (options.apiToken) {
+    return resin.token.set(options.apiToken)
+  }
+  return Promise.resolve()
```
pimterry: You don't need this line - doing nothing (implicitly `return undefined`) should work just as well.

bin/resin-preload
```diff
@@ -132,6 +131,12 @@ buildOutputStream.pipe(process.stdout)
 
 streamToPromise(buildOutputStream)
 .then(() => {
+  if (options.apiToken) {
+    return resin.token.set(options.apiToken)
```
pimterry: This will work I think, but will also set your user-wide token, updating the token for the CLI, which might cause some unexpected behaviour for some users, and is an odd side effect.

You can fix that by specifying a different data directory when you create the SDK, by passing `dataDirectory: <tmpDir>` in `resinSdk({ })` above. You'll need some library to give you a new tmp dir each time though, and to clean it up after too.

Is the behaviour where an existing CLI token is used automatically desirable? If it's not, you can just always use a temp dir. If it is, though you'll want to use a temp dir only if `options.apiToken` is provided, and stick with the default otherwise (so any existing tokens there get found).

zvin: Fixed, I used a temporary directory for all cases.
If someone wants to use the resin-cli token, he can use `resin-cli preload`.

bin/resin-preload
```diff
@@ -88,6 +92,21 @@ if (argv.includes('--version') || argv.includes('-v')) {
   process.exit(0)
 }
 
+const getResinSdk = (options) => {
+  // Creates a temporary directory for resin sdk so it won't replace any existing token.
+  const resinSdkOptions = { apiKey: options.apiKey, apiUrl: options.apiHost }
+  return tmp.dirAsync({ dir: os.tmpdir(), unsafeCleanup: true })
```
pimterry: Doesn't break anything, but I don't think you need the `dir` argument here - `os.tmpdir()` is the default.

zvin: Good catch, removed.

bin/resin-preload
```diff
@@ -8,9 +8,12 @@ const info = require('../package.json')
 const Promise = require('bluebird')
 const streamToPromise = require('stream-to-promise')
 const resinSdk = require('resin-sdk')
+const tmp = Promise.promisifyAll(require('tmp'))
 
 const docker = new Docker({Promise})
 
+tmp.setGracefulCleanup()
```
jhermsmeier: I would not recommend using `tmp`, as its cleanup doesn't actually work in most cases (as we found out in drivelist), see https://github.com/resin-io-modules/drivelist/pull/191 and https://github.com/resin-io/etcher/issues/1571

Issue #92, opened 2017-08-21
TypeError: process.stdin.setRawMode is not a function


zvin:
https://github.com/resin-io/resin-cli/issues/637

PR #93, opened 2017-08-21
Don't call stdin.setRawMode if it doesn't exist.

zvin:
Connects-To: #92
Connects-To: https://github.com/resin-io/resin-cli/issues/637

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #92


lib/preload.js
```diff
@@ -100,14 +100,18 @@ const runContainer = (docker, container) => {
...
+    }
   }
-  process.stdin.resume()
   // Set raw mode to be able to intercept CTRL-C
-  process.stdin.setRawMode(true)
-  process.stdin.on('data', killContainer)
+  if (process.stdin.setRawMode) {
+    process.stdin.resume()
+    process.stdin.setRawMode(true)
+    process.stdin.on('data', killContainer)
```
pimterry: Are you sure `on('data', ...` won't work here? This will apply to anybody running this in Windows, or over SSH connections, and in both cases I imagine users will expect ctrl-c to work as normal.

zvin: When I tried, I couldn't catch CTRL-C without setting raw mode.

PR #95, opened 2017-08-22
Add progress bars and spinners.

zvin:
Connects-To: #88

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #88


lib/preload.js
```diff
@@ -251,11 +265,19 @@ preload.CONTAINER_NAME = 'resin-image-preloader'
 
 /**
  * Build the preloader docker image
+ * @param {Object} docker - dockerode instance
+ * @param {Object} visuals - resin-cli-visuals (optional)
  * @returns {Stream}
  */
-preload.build = (docker) => {
+preload.build = (docker, visuals) => {
```
pimterry: Why doesn't this module just depend on & require `resin-cli-visuals` itself?

lib/preload.js
```diff
@@ -251,11 +265,19 @@ preload.CONTAINER_NAME = 'resin-image-preloader'
...
   const stream = new streamModule.PassThrough()
   const files = ['Dockerfile', 'requirements.txt', 'src/preload.py']
+  let progress
+  let matches
```
pimterry: Is there a good reason to hoist this `matches` here? You could declare it inside its `if` below and it should work fine.

lib/preload.js
zvin: There are 2 reasons:
1. the standalone mode (bin/resin-preload) doesn't use/need visuals;
2. To avoid having 2 copies of resin-cli-visuals if the versions diverge

lib/preload.js
zvin: moved

lib/preload.js
pimterry: For 1, personally I wouldn't worry. I agree it is slightly messy, but it doesn't actually cause any problems, and it's a pretty tiny module.

For 2, this is actually a big negative: if the versions diverge, that means this module is now going to get given an instance of visuals that has breaking changes in it, and may not behave in the way it expects. If we release a version of `resin-cli-visuals` that renames `visuals.Progress` to `visuals.ProgressBar`, this code will quietly and unexpectedly implode, because it doesn't say or check anywhere which version of visuals it's expecting. If we make this module depend on visuals directly instead, then we would end up with two copies, but that's a tiny problem, and everything keeps working in the meantime.

Does it cause problems if the standalone mode has nice visuals, or is it just that it's not useful? If it's not going to break anything, I think I'd just depend on it and enable visuals all the time: it solves the above problem, and nicely simplifies the code by letting you remove the `if (visuals)` everywhere too.

zvin: The standalone mode has no visuals.
It is deprecated and won't get any new feature (like visuals).
I'm just keeping it functional until `preload` through `resin-cli` is well tested.
I'll probably remove it next month.

I'll make the changes to require `resin-cli-visuals` in resin-preload.

zvin: done

lib/preload.js
```diff
@@ -196,17 +198,24 @@ const getLayerSize = (registryHost, token, imageRepo, blobSum) => {
 }
 
 const getContainerSize = (apiHost, apiToken, registryHost, imageRepo) => {
-  console.log('Fetching container size for', imageRepo)
+  const message = `Fetching container size for ${imageRepo}`
+  let progress
+  progress = new visuals.Progress(message)
```
zvin: done

lib/preload.js
```diff
@@ -251,11 +260,15 @@ preload.CONTAINER_NAME = 'resin-image-preloader'
...
 /**
  * Build the preloader docker image
+ * @param {Object} docker - dockerode instance
  * @returns {Stream}
  */
 preload.build = (docker) => {
   const stream = new streamModule.PassThrough()
   const files = ['Dockerfile', 'requirements.txt', 'src/preload.py']
+  let progress
+  progress = new visuals.Progress('Building Docker preloader image.')
```
zvin: done

lib/preload.js
```diff
@@ -196,17 +198,26 @@ const getLayerSize = (registryHost, token, imageRepo, blobSum) => {
...
+        progress.update({ percentage: finished / size * 100 })
+      }
+      updateProgress()
       return Promise.map(
         _.entries(layersCount),
         (layer) => {
           return getLayerSize(registryHost, token, imageRepo, layer[0])
           .then((layerSize) => {
+            finished += 1
+            updateProgress()
```
lurch: Doesn't matter, but I guess you could have initialised `let finished = 0` and then done `finished += 1` as the first line inside `updateProgress` ?

Issue #94, opened 2017-08-21
Handle preloading on devices that have the `overlay2` storage driver.


zvin:
The driver to use may be in `/lib/systemd/system/docker.service`.
If not, add an option to resin-preload.

PR #96, opened 2017-08-23
Use the docker storage driver from /<rootfs>/lib/systemd/system/docker.service ExecStart key.

zvin:
With this change we no longer need to extract the resinOS version and make decisions based on it.

overlay2 storage driver works (nvidia jetson devices for example)

This should also solve #89 as we wait for dockerd to exit.

Connects-To: #94

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #94


CameronDiver:
@zvin could you provide a little more explanation about why we need this change please?

zvin:
@CameronDiver This is for #94 
Preload needs to work with images that use the overlay2 Docker storage driver like the nvidia jetson tx2.

The storage driver Docker uses is in the `/lib/systemd/system/docker.service` file on the root partition.
It is a systemd unit file with a line like this:
```
ExecStart=/usr/bin/dockerd --log-driver=journald -s overlay2 -H fd:// --dns 172.17.0.1
```

We open this file, see what's after `-s` and use it to start dockerd.

Before this PR, we were getting the resinOS version and using `btrfs` if version was < 2 and `aufs` else.

src/preload.py
```diff
@@ -389,6 +377,38 @@ def list_images(image):
...
 
 
+def get_docker_init_file_content(image):
+    with mount_context_manager(image, 2) as mountpoint:
+        path = os.path.join(
+            mountpoint,
+            'lib',
+            'systemd',
+            'system',
+            'docker.service',
```
pimterry: Presumably this works for ResinOS, but it's not guaranteed to work for images more generally. Docker's recommended location for setting like this is actually `/etc/docker/daemon.json`, and anybody starting docker outside systemd could be passing in these args absolutely elsewhere.

Is that a concern? What about with host apps in the future? Is there any more reliable way we could work out the filesystem that's being used?

Looking at some of my devices, I've got a `/var/lib/docker/aufs` folder on new ResinOS 2 devices, and I've got `/var/lib/docker/overlay2` on my local machine (which uses overlay2). Is checking for the existence of those practical, and any more reliable?

zvin: This script is for preloading resinOS images, I don't think we care about anything else.
I'm not sure that checking for folders existence would be more reliable (at least for resinOS).
To handle other cases we could add a `--force-storage-driver <driver>` option.

zvin: @pimterry this is definitely only meant for resinOS images.
One of the first things preload does is to find a config.json or device-type.json file.
Before this PR the script was looking for /etc/os-release.
It also assumes that the root partition in the 2nd and that the docker data is on the 6th partition.

pimterry: Ok, that's fair enough then, as long as we're happy with that constraint :+1:

Issue #98, opened 2017-08-24
NPM package contains `__pycache__`


imrehg:
The NPM package (at least 3.1.0), has precompiled python binary cached there, which probably shouldn't be there, especially because it's out of date (Jul 18, while the latest version of the python script it caches is Aug 22): `src/__pycache__`.

Might need to look at the packaging, and make sure that is not shipped?

zvin:
Good catch @imrehg , removing ASAP.

PR #99, opened 2017-08-24
Add a files section to package.json.

zvin:
Connects-To: #98

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #98


package.json
```diff
@@ -12,6 +12,13 @@
     "container"
   ],
   "main": "lib/preload.js",
+  "files": [
+    "bin/resin-preload",
+    "lib/preload.js",
+    "src/preload.py",
+    "Dockerfile",
+    "requirements.txt"
+  ],
```
jhermsmeier: Instead of whitelisting things explicitly, I think it'd be less error prone to just add things like `__pycache__` to the `.npmignore`

PR #97, opened 2017-08-24
Add flag for ignoring mismatching device types

svenschwermer:

PR #100, opened 2017-08-25
Fix js style (standard js)

zvin:
Change-Type: patch


zvin:
This one got through because #97 was merged manually.

PR #117, opened 2017-10-04
Allow preloading of Edison zip archives.

zvin:
* Also use registry2 host from resin-sdk
 * Refactor allowing easy creation of progress bars and spinners

Connects-To: #116

Change-Type: major
---- Autogenerated Waffleboard Connection: Connects to #116


lib/preload.js
```diff
@@ -66,10 +108,397 @@ const bindMount = (source, target) => {
...
 }
 
-const createContainer = (docker, options) => {
-  const mounts = [bindMount(options.image, DISK_IMAGE_PATH_IN_DOCKER)]
-  if (options.splashImage) {
-    mounts.push(bindMount(options.splashImage, SPLASH_IMAGE_PATH_IN_DOCKER))
+const zipContainsFiles = (archive, files) => {
+  // FIXME: read from the zip directory instead of readin the whole file
+  // This is broken in unzipper for now: an invalid zip file will throw an
+  // uncaught error event.
```
CameronDiver: Is this something that we can use and then catch the error?

zvin: It will emit an error here https://github.com/ZJONSSON/node-unzipper/blob/master/lib/PullStream.js#L66 .
We don't have access to this stream with `node-unzipper`'s api so we can't handle it

zvin: I mean if we use
```javascript
return unzipper.Open.file(archive)
  .then(function(d) {
    ...
  })
```

zvin: and `node-unzipper` soesn't handle it either.

PR #119, opened 2017-10-06
Don't use Array.includes, it is not supported in node4.

zvin:
Connects-To: #118

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #118


pimterry:
Do we care about Node 4? The CLI only supports 6+: https://github.com/resin-io/resin-cli/blob/master/package.json#L31

zvin:
@pimterry resin-preload can work without resin-cli and one of our customers needs node 4 support.

PR #124, opened 2017-10-13
Get R_OK and W_OK constants from wherever they're available

pcarranzav:
…to keep supporting node v4 for the time being

`fs.constants` is not present on node v4, so in that case we use fs.R_OK and fs.W_OK directly.

Change-Type: patch


lib/preload.js
```diff
@@ -19,6 +19,9 @@ const getFolderSizeAsync = Promise.promisify(require('get-folder-size'))
 
 const preload = module.exports
 
+const R_OK = fs.constants != null ? fs.constants.R_OK : fs.R_OK
+const W_OK = fs.constants != null ? fs.constants.W_OK : fs.W_OK
```
zvin: can we use 
```javascript
const R_OK = _.isUndefined(fs.constants) ? fs.constants.R_OK : fs.R_OK
const W_OK = _.isUndefined(fs.constants) ? fs.constants.W_OK : fs.W_OK
```
please ?

PR #125, opened 2017-10-17
Fix the logic that defines where we get R_OK and W_OK constants for n…

pcarranzav:
…ode 4 support

The _.isUndefined causes the logic to be inverted, so we need to use fs.R_OK if the condition is true.

Change-Type: patch


zvin:
Sorry about that @pcarranzav

pcarranzav:
no worries @zvin - I've actually done a few similar ones recently :)
would you ping me when it's published on npm?

zvin:
@pcarranzav it's published now: https://www.npmjs.com/package/resin-preload

pcarranzav:
Awesome, thanks!

PR #126, opened 2017-10-18
Support resinOS >= 2.6

zvin:
---- Autogenerated Waffleboard Connection: Connects to #112


src/preload.py
```diff
@@ -427,15 +427,38 @@ def list_images(detect_flasher_type_images):
...
+
+def _find_docker_overlay2_root(mountpoint):
+    path = os.path.join(mountpoint, "docker", "overlay2")
+    if os.path.exists(path) and os.path.isdir(path):
+        for folder in os.listdir(path):
+            folder_path = os.path.join(path, folder, "diff")
+            if os.path.exists(folder_path) and os.path.isdir(folder_path):
+                files = os.listdir(folder_path)
+                if any(x for x in files if not x.startswith(".")):
+                    return folder_path
```
jhermsmeier: As the above two functions pretty much look the same at first glance, can you add some comments in the places they differ, to make it obvious and quicker to read?

bin/resin-preload
```diff
@@ -160,6 +160,7 @@ const spinnerHandler = (event) => {
   if (event.action === 'start') {
     spinner.start()
   } else {
+    console.log()
```
jhermsmeier: What is this for?

zvin: It's the same as https://github.com/resin-io/resin-cli/pull/687#discussion_r144904431

src/preload.py
zvin: Comments added

bin/resin-preload
jhermsmeier: Ah okay – can you also add a short comment about that here?

zvin: Added

jhermsmeier:
Looks like you'll need to rebase this against master

src/preload.py
CameronDiver: Even better, could you gather the repeated code into a worker function and call into that from these two?

zvin: This will only make it unreadable, let's keep things simple.

zvin: I factorized the "find a non empty folder" part

src/preload.py
```diff
@@ -427,15 +427,45 @@ def list_images(detect_flasher_type_images):
...
+
+def _find_docker_overlay2_root(mountpoint):
+    # We're looking for a /docker/overlay2/<xxxxxxxxxxxxx>/diff folder with
+    # some files not starting with a '.'
+    path = os.path.join(mountpoint, "docker", "overlay2")
+    if os.path.exists(path) and os.path.isdir(path):
+        for folder in os.listdir(path):
+            folder_path = os.path.join(path, folder, "diff")
+            if _is_non_empty_folder(folder_path):
+                return folder_path
```
CameronDiver: You can replace these with:
```python
# You probably want a better function name
def find_non_empty_path(path, childDir = ''):
    if os.path.exists(path) and os.path.isdir(path):
        for folder in os.listdir(path):
            folder_path = os.path.join(path, folder, childDir) if childDir != '' else os.path.join(path, folder)
            if _is_non_empty_folder(folder_path):
                return folder_path

def _find_docker_aufs_root(mountpoint):
    path = os.path.join(mountpoint, 'docker', 'aufs', 'diff')
    return find_non_empty_path(path)

def _find_docker_overlay2_root(mountpoint):
    path = os.path.join(mountpoint, 'docker', 'overlay2')
    return find_non_empty_path(path, 'diff')
```
It's untested but it cleans up this code block a bit.

src/preload.py
CameronDiver: Oh I didn't see these comments for some reason. Anyway I added some different code below, check it out, if you still feel that it's more unreadable feel free to merge. I personally think it's a bit nicer though.

Issue #127, opened 2017-10-19
Some images have important data outside of partitions or partition tables.


zvin:
Like Variscite DART-6UL for example.

It's a flashed type device.

When preloading this type of image, the preload script needs to recreate a disk image in order to be able to move partitions after the root partition (which needs to be resized).
This process creates the new partition table using sfdisk and then copies the partitions from the old image to the new one.
This means that data outside of partitions is not copied.

As https://community.nxp.com/docs/DOC-95015 says,
> U-boot should reside at offset 1024B of your SD card.

We need to copy data between the partition table and the first partition to the new image.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_38s1)

lurch:
ping @jhermsmeier and @jviotti - this is obviously something to be aware of in Etcher too, when we add the "dynamic blockmap" code and ignore any non-partitioned diskspace.

PR #128, opened 2017-10-19
Copy data between the MBR and the first partition when resizing a preloader image.

zvin:
Connects-To: #127

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #127


src/preload.py
```diff
@@ -324,6 +325,14 @@ def resize_rootfs(additional_space):
     sfdisk(tmp.name, _in=new_layout)
     offsets_and_sizes = get_offsets_and_sizes(image, "s")
     copy = partial(ddd, _if=image, of=tmp.name, bs=SECTOR_SIZE, conv="notrunc")
+    # Copy data between the MBR and the first partition (some devices rely on
+    # the bootloader being there, like the Variscite DART-6UL)
+    mbr_size_in_sectors = MBR_SIZE // SECTOR_SIZE
```
CameronDiver: I could be wrong, but don't you want to `ceil()` this number, rather than `floor()`? If it takes 0.5 of another sector, surely we need to copy that sector as well? Or am I missing something?

zvin: This relies on the fact that MBR_SIZE == SECTOR_SIZE == 512.
I will update this in the upcoming GPT support PR where the `dd`'s `bs` will be increased and the `seek`, `skip` and `count` will use bytes.

CameronDiver: Please could you add a comment detailing this then, that could definitely catch somebody out.

imrehg:
Just going to test this and report back.

src/preload.py
```diff
@@ -329,6 +330,16 @@ def resize_rootfs(additional_space):
     sfdisk(tmp.name, _in=new_layout)
     offsets_and_sizes = get_offsets_and_sizes(image, "s")
     copy = partial(ddd, _if=image, of=tmp.name, bs=SECTOR_SIZE, conv="notrunc")
+    # Copy data between the MBR and the first partition (some devices rely on
```
lurch: I wonder if this would be clearer if reworded like "Copy across any data that's located between the MBR and the first partition ..." ?
I initially misinterpreted the line above as "Copy data from the MBR to the first partition ..." and had to `man dd` to work out what the arguments below are actually doing ;-)

zvin: Sure, comment updated.

Issue #134, opened 2017-10-24
Fix incompatibility with docker 1.12.6 (api version 1.24)


zvin:
https://app.frontapp.com/open/cnv_9wa3w9

PR #135, opened 2017-10-24
Fix incompatibility with Docker 1.12.6 (api version 1.24)

zvin:
Prior to api 1.25 (~docker 1.12) bind mounts were going to `HostConfig.Binds` when creating a container.
In api 1.25 and up they go to `HostConfig.Mounts`.
Also, they changed from being a string like
`source:destination`
to an object like
```javascript
{
  Source: source,
  Target: target,
  Type: 'bind',
  Consistency: 'delegated'  // this is only needed to speed things up on macOS
}
```

This PR detects the docker api version and uses the right field.

Connects-To: #134 

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #134


CameronDiver:
@zvin what is the incompatibility which this is tackling and how are you tackling it? Could you add this to the PR description for reference for future us please?

zvin:
@CameronDiver added.

package.json
```diff
@@ -22,8 +22,14 @@
...
     "archiver": "^2.0.3",
     "bluebird": "^3.5.0",
+    "compare-versions": "^3.1.0",
     "docker-progress": "^2.6.3",
     "dockerode": "^2.5.0",
+    "eslint-plugin-import": "^2.8.0",
+    "eslint-plugin-node": "^5.2.1",
+    "eslint-plugin-promise": "^3.6.0",
+    "eslint-plugin-react": "^7.4.0",
+    "eslint-plugin-standard": "^3.0.1",
```
jhermsmeier: Where did all these `eslint-plugin-*` dependencies come from? 🐣 
Have they been added intentionally, or is this material for / from another PR?

zvin: Good catch, that's an error. Removed.

Issue #120, opened 2017-10-12
Add GPT images support


zvin:
For preloading `jetson-tx2` images for example.

PR #138, opened 2017-10-26
Preload GPT disk images, remove --dont-detect-flasher-type-images option

zvin:
* use partition labels instead of hardcoding partition numbers
 * better flasher image detection, remove the no longer needed --dont-detect-flasher-type-images option
 * fix preloading resinOS 1.8 flasher images

Connects-To: #120

Change-Type: major
---- Autogenerated Waffleboard Connection: Connects to #120


src/preload.py
```diff
@@ -42,116 +45,371 @@
...
+                '(BTRFS Filesystem label|label:|volume name) "(.*)"',
+                out,
+            )
+            if match is not None:
+                return match.groups()[1].strip()
+
+    def set_parent(self, parent):
+        # For logical partitions on MBR disks we store the parent extended
+        # partition
+        assert self.partition_table.label == "dos"
```
CameronDiver: Do we want to leave an assert in production code?

src/preload.py
```diff
@@ -505,40 +673,56 @@ def get_docker_storage_driver(image=None):
...
-        part = get_partition("root")
-        log.info(
-            "This is a flasher image, preloading into /opt/{} on the 2nd "
-            "partition of {}".format(inner_image, part["image"])
-        )
-        resize_rootfs(additional_space)
-        with mount_context_manager("root") as mountpoint:
-            image = os.path.join(mountpoint, "opt", inner_image)
-            preload(additional_space, app_data, image)
+    additional_bytes = round_to_sector_size(ceil(int(container_size) * 1.1))
```
CameronDiver: What is the `1.1` magic number?

src/preload.py
zvin: Yes, preload.py has several asserts.
I think it might help debugging.
Without asserts, weird errors will happen later.

src/preload.py
zvin: It's the 110% from the comment on the line above:
>Size will be increased by 110% of the container size

CameronDiver: Ah yeah, the diff made that quite hard to see :)

src/preload.py
CameronDiver: Gotcha, just wanted to make sure it was being considered.

imrehg:
Tested so far:

* Jetson TX2 2.7.4+rev1: seems to be working fine

Tested on RPi3 and BBB 1.8.0, and so far no success, not sure if they should work, based on the commit message.

src/preload.py
```diff
@@ -42,116 +45,371 @@
...
+        return mount_context_manager(
+            self.image,
+            self.start_bytes,
+            self.size_bytes,
+        )
+
+    def str(self):
+        return "partition n°{} of {}".format(self.number, self.image)
+
+    def _resize_last_partition_of_disk_image(self, additional_bytes):
```
lurch: The fact that `_resize_last_partition_of_disk_image` and `_resize_partition_on_disk_image` both have the same `(self, additional_bytes)` method-signature looks a little confusing. Not that it really matters, but would these methods make "more sense" on a `Disk` class, rather than a `Partition` class?

zvin: This is because they were part of Partition.resize().
This method got too large and I had to split it.
I still find them too long but couldn't find a way to split them more without reducing readability.

I don't understand the second comment. We don't have a Disk class (but we have PartitionTable). We want to resize the partition so Partition.resize seems to be the good place. Note that a Partition knows the PartitionTable it belongs to.

lurch: My apologies - I was commenting while tired and got confused ;-)
"Note that a Partition knows the PartitionTable it belongs to" is the bit I didn't realise.

zvin: No problem, feedback is always appreciated.

src/preload.py
```diff
@@ -42,116 +45,371 @@
...
+
+
+def get_filesystem(device):
+    line = fsck("-N", device).stdout.decode("utf8").strip().split("\n")[1]
+    return line.rsplit(" ", 2)[-2].split(".")[1]
+
+
+def expand_filesystem(partition):
+    with partition.losetup_context_manager() as device:
+        # Detects the partition filesystem (ext{2,3,4} or btrfs) and uses the
```
lurch: Should this raise some kind of error if the partition _isn't_ using "(ext{2,3,4} or btrfs)" filesystem?

zvin: Good idea but this PR is already merged, I'll try to think of it for the next one.

PR #147, opened 2017-11-17
Wait for loop device to be detached before killing the container

zvin:
Connects-To: #146

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #146


lib/preload.js
```diff
@@ -437,16 +434,19 @@ class Preloader extends EventEmitter {
...
       return innerDockerProgress.pull(appData.imageId, (event) => {
         // Emit progress events while pulling
-        percentage = event.percentage
         this._progress(pullingProgressName, event.percentage)
       })
     })
     .then(() => {
-      if (percentage !== 100) {
-        return
-      }
```
pimterry: Ok, I don't understand a lot of the context of this. Presumably this code used to not signal 'done' if percentage !== 100 for some reason before though. Why was that, and why is it no longer relevant?

zvin: @pimterry it's dead code. It was introduced by https://github.com/resin-io/resin-preload/commit/75addc2a291674d92250da495cc56f3db167026a .
This never happens.
Even if it happened, we should still signal the python process that we're done pulling.

lib/preload.js
```diff
@@ -437,16 +434,19 @@ class Preloader extends EventEmitter {
...
       // Signal that we're done to the Python script.
       this.stdin.write('\n')
+      // Wait for the script to unmount the data partition
+      return new Promise((resolve, reject) => {
+        this.stdout.on('error', reject)
+        this.stdout.once('data', resolve)
```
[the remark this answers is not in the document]
zvin: updated

lib/preload.js
```diff
@@ -437,16 +434,19 @@ class Preloader extends EventEmitter {
...
     })
     .then(() => {
-      if (percentage !== 100) {
-        return
-      }
       // Signal that we're done to the Python script.
       this.stdin.write('\n')
+      // Wait for the script to unmount the data partition
+      return new Promise((resolve, reject) => {
+        this.stdout.once('error', reject)
```
jhermsmeier: What errors are you expecting to handle here? Was this meant to be `this.stderr.once('data', reject)` or is this meant to handle cases where the child process errors/exits, or is this something docker.modem does (emitting errors on the stream)?

jhermsmeier: PS: Sorry I'm a bit late to the party

zvin: I'm not expecting an error here, I've just put it there so if an error happens it will be easier to debug.
`data` on `stderr` is expected, we use it as a debug output.

jhermsmeier: Cool, thanks! :)

PR #149, opened 2017-12-12
Preload images that use Balena instead of Docker

zvin:
Connects-To: #148

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #148


src/preload.py
```diff
@@ -633,30 +633,43 @@ def find_non_empty_folder_in_path(path, child_dir=""):
...
 def find_docker_aufs_root(mountpoint):
-    # We're looking for a /docker/aufs/diff/<xxxxxxxxxxxxx>/ folder with some
+    # We're looking for a /<docker|balena>/aufs/diff/<xxxxxxxxxxxxx>/ folder with some
     # files not starting with a '.'
-    path = os.path.join(mountpoint, "docker", "aufs", "diff")
-    return find_non_empty_folder_in_path(path)
+    for name in ("docker", "balena"):
+        path = os.path.join(mountpoint, name)
+        if os.path.isdir(path):
+            path = os.path.join(path, "aufs", "diff")
```
lurch: Only a minor tweak, but I believe 
```python
path = os.path.join(mountpoint, name)
if os.path.isdir(path):
     path = os.path.join(path, "aufs", "diff")
```
 could be collapsed to just:
```python
path = os.path.join(mountpoint, name, "aufs", "diff")
if os.path.isdir(path):
```

Hmmm, although of course that _would_ behave slightly different to the code you're using if we have both `/docker/foo` and `/balena/aufs/diff/foo` directories at the same time ;-)  I don't know enough about resinos to know if that'd ever be the case.

(and the same point also applies to the two separate `os.path.join` calls used in `find_docker_overlay2_root`)

src/preload.py
```diff
@@ -633,30 +633,43 @@ def find_non_empty_folder_in_path(path, child_dir=""):
...
+        path = os.path.join(mountpoint, name)
+        if os.path.isdir(path):
+            path = os.path.join(path, "overlay2")
+            return find_non_empty_folder_in_path(path, "diff")
 
 
-def get_docker_init_file_content(image=None):
+def get_docker_service_file_path(folder):
+    for name in ("docker", "balena"):
+        fpath = os.path.join(folder, "lib/systemd/system", name + ".service")
```
lurch: This _could_ be
```
fpath = os.path.join(folder, "lib", "systemd", "system", name + ".service")
```

lurch: Hmmm, if `find_docker_aufs_root` or `find_docker_overlay2_root` has already identified whether we're on a "docker system" or a "balena system", will we _ever_ have a mis-match of the two? I.e. could we ever have:
```
/docker/overlay2/<xxxxxxxxxxxxx>/diff/lib/systemd/system/balena.service
```
or
```
/balena/aufs/diff/<xxxxxxxxxxxxx>/lib/systemd/system/docker.service
```
?

src/preload.py
```diff
@@ -633,30 +633,43 @@ def find_non_empty_folder_in_path(path, child_dir=""):
...
+        if os.path.exists(fpath):
+            return fpath
+
+
+def get_docker_service_file_content(image=None):
     docker_service_file_path = "lib/systemd/system/docker.service"
```
lurch: Looks like this `docker_service_file_path` variable is now redundant

PR #153, opened 2018-01-10
Bind mount /dev on linux hosts

zvin:
This fixes the losetup problem on Fedora 25

Connects-To: #152

Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #152


lurch:
If the container now has access to the host's `/dev`, does that have any security implications? Or does no user-code get run during the preloading process?

zvin:
@lurch we only use /dev/loopX files in the container while preloading, no user code is ran.

PR #155, opened 2018-01-25
Multicontainer preload [DO NOT MERGE]

zvin:
* use the new multicontainer resin-sdk
 * resize disk image only if it lacks space
 * preload multicontainer apps if the image supervisor version is >= 7.0.0
 * simplify api
 * drop request and request-promise requirements
 * get docker images size from the api, no longer query the registry

Connects-To: #154

Change-Type: major
---- Autogenerated Waffleboard Connection: Connects to #154


bin/resin-preload
```diff
@@ -203,33 +202,16 @@ getResinSdk(options)
...
     .then(() => {
-      if (preloader.application.build.length === 0) {
-        reject(new ExitError('This application has no successful builds'))
-      }
-      return preloader.getDeviceTypeAndPreloadedBuilds()
-    })
-    .then((info) => {
-      if (options.dontCheckDeviceType === false && info.device_type !== preloader.application.device_type) {
-        const msg = `Application device type (${preloader.application.device_type}) and image device type (${info.device_type}) do not match.`
+      const msg = preloader.cantPreload()
```
dfunckt: `cantPreload` is a slightly weird name for a function that returns an error message if some conditions aren't met. What this actually does is validate that preloading can be performed, so in my opinion it should named something like `ensureCanPreload` that *throws errors* if any of the conditions it checks aren't met.

bin/resin-preload
```diff
@@ -203,33 +202,16 @@ getResinSdk(options)
...
+      if (msg) {
         reject(new ExitError(msg))
+      } else {
+	return preloader.preload()
```
dfunckt: There's some issue here with indentation.

bin/resin-preload
```diff
@@ -203,33 +202,16 @@ getResinSdk(options)
...
-      const builds = info.preloaded_builds.map((build) => {
-        return build.slice(-preload.BUILD_HASH_LENGTH)
-      })
-      if (builds.indexOf(options.commit || preloader.application.commit) !== -1) {
-        reject(new ExitError('This build is already preloaded in this image.'))
-      }
-      return preloader.preload()
-      .then(resolve)
     })
+    .then(resolve)
```
dfunckt: You can resolve the promise earlier and remove these two lines (`then` and `catch` below`).

lib/preload.js
```diff
@@ -127,8 +125,81 @@ const isEdisonZipArchive = (file) => {
...
+    return docker.createContainer(containerOptions)
+  })
+}
+
+const isReadWriteAccessibleFile = (image) => {
+  return Promise.all([
+    fs.accessAsync(image, R_OK | W_OK),
+    fs.statAsync(image)
+  ])
+  .spread((_, stats) => {
```
dfunckt: `Promise.join` is slightly better than `all+spread` when starting a promise chain.

lib/preload.js
```diff
@@ -127,8 +125,81 @@ const isEdisonZipArchive = (file) => {
...
+    return stats.isFile()
+  })
+  .catchReturn(false)
+}
+
+const checkImage = (image) => {
+  return isReadWriteAccessibleFile(image)
+  .then((ok) => {
+    if (!ok) {
+      throw new errors.ResinError('The image must be a read / write accessible file')
```
dfunckt: *read / write* -> *read/write*

lib/preload.js
```diff
@@ -127,8 +125,81 @@ const isEdisonZipArchive = (file) => {
...
+    }
+  })
+}
+
 class Preloader extends EventEmitter {
-  constructor (resin, docker, appId, commit, image, splashImage, proxy) {
+  constructor (resin, docker, appId, commit, image, splashImage, proxy, dontCheckDeviceType) {
```
dfunckt: `dontCheckDeviceType` reads slightly weird. I think `shouldSkipDeviceTypeCheck` (or `skipDeviceTypeCheck` is you prefer brevity) is more appropriate, but I'm nit picking.

lib/preload.js
```diff
@@ -242,10 +319,172 @@ class Preloader extends EventEmitter {
...
+        }
+        this.emit('error', error)
+      }
+    })
+  }
+
+  _runCommand (command, parameters) {
+    return new Promise((resolve, reject) => {
+      this.stdout.once('error', reject)
+      this.stdout.once('data', (data) => {
```
dfunckt: Are you sure `once` is enough to gather all output? It seems more appropriate to register with `.on` and remove the listener when `data === null`.

lib/preload.js
zvin: done

lib/preload.js
zvin: I'll keep it as it is as it is the name of the option (--dont-check-device-type).

bin/resin-preload
zvin: fixed

lib/preload.js
zvin: `once` enough here. The responses are always JSON encode strings with no newlines.
I'd prefer not changing this in this PR.

lib/preload.js
zvin: done

bin/resin-preload
zvin: I don't think I can do that, this code is inside a `new Promise((resolve, reject) => {...`.

bin/resin-preload
zvin: I didn't want this function to throw to simplify the resin-preload api.

zvin: Finally I updated this to be _ensureCanPreload: a method that throws a ResinError if something is wrong.

PR #158, opened 2018-02-06
Preserve GRUB in MBR when resizing

zvin:
Connects-To: #157

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #157


src/preload.py
```diff
@@ -31,9 +31,11 @@
 
 IMAGE = "/img/resin.img"
 
+# In bytes:
 SECTOR_SIZE = 512
 MBR_SIZE = 512
 GPT_SIZE = SECTOR_SIZE * 34
+GRUB_SIZE = 446
```
lurch: To be pedantic, perhaps this should be `MBR_BOOTSTRAP_CODE_SIZE` ?

Looking at https://en.wikipedia.org/wiki/Master_boot_record#Sector_layout I guess if we copy the first 446 bytes, we'll also be copying the disk-signature. ISTR that Linux uses the disk-signature to generate/store the disk's UUID values, so I guess this means we'd be creating new images with identical UUIDs?
I've got no idea if that's something we want to be doing or not, just pointing ti out ;-)

src/preload.py
```diff
@@ -288,6 +290,8 @@ def _resize_partition_on_disk_image(self, additional_bytes):
         # partition (some devices rely on the bootloader being there, like the
         # Variscite DART-6UL)
         if self.partition_table.label == "dos":
```
lurch: Do GPT-partitioned disks also need to have this bootstrap-code area copied? (I googled and found https://askubuntu.com/questions/657179/where-is-grub-installed-on-a-gpt-disk )

src/preload.py
zvin: Good catch for the name, i amende my commit.
About the uuid generation, resin-preload is not supposed to update the disk uuid so that's fine.

src/preload.py
zvin: I don't think we have any image that uses the "BIOS-mode booting from GPT" but as there is a protective MBR in the beginning of a GPT partitionned disk I think we can safely copy the first 446 bytes in all cases.
Sounds good @lurch ?

zvin: code updated to copy the first 446 bytes in all cases

PR #175, opened 2018-05-30
Allow devices to be pinned following a preload provision

CameronDiver:
Change-type: minor
Closes: #174
Signed-off-by: Cameron Diver <cameron@resin.io>


lib/preload.js
```diff
@@ -633,7 +634,9 @@ class Preloader extends EventEmitter {
         return _.merge({}, _.omit(value, [ 'environment', 'image' ]), { appId, env: value.environment, imageId: value.image })
       })
     } else {
-      return this.state.local
+      let data = _.cloneDeep(this.state.local)
+      _.merge(data, { pinDevice: this.pinDevice })
+      return data
```
zvin: NIT:
```javascript
return _.merge({}, this.state.local, { pinDevice: this.pinDevice })
```

CameronDiver: Ah yeah, I was trying to avoid changing `this.state.local`, but your way is much nicer :)

PR #181, opened 2018-08-20
Check device architecture instead of device type

zvin:
Change-type: major
Signed-off-by: Alexis Svinartchouk <alexis@resin.io>


lib/preload.js
```diff
@@ -498,6 +507,13 @@ class Preloader extends EventEmitter {
     return `${count} ${thing}${(count !== 1) ? 's' : ''}`
   }
 
+  _deviceTypeArch (deviceType) {
+    return _.find(this.deviceTypes, (dt) => {
+      return (dt.slug === deviceType)
+    })
+    .arch
```
CameronDiver: This doesn't handle not finding the device type, and the error shown wouldn't be that great (`cannot find arch of undefined`)

zvin: Good catch, updated.

PR #182, opened 2018-08-21
Pinning information should also be available for the resin-image-flasher

telphan:
Change-type: patch
Signed-off-by: Theodor Gherzan <theodor@resin.io>


src/preload.py
```diff
@@ -515,6 +515,15 @@ def docker_context_manager(storage_driver, mountpoint):
     if kv_file_existed:
         write_file(local_kv_db_path, local_kv_db_content)
 
+def write_resin_device_pinning(app_data, output):
+    """Create resin-device-pinnnig.json to hold pinning information"""
+    if(not app_data['pinDevice']):
```
zvin: No need for parenthesis around conditions.
```python
def write_resin_device_pinning(app_data, output):
    if app_data.get("pinDevice") is None:
       return
    apps = app_data.get("apps", {})
    if len(apps) != 1:
        raise Exception("Malformed apps.json")
    with open(output, "w") as f:
        f.write("RELEASE_ID={}".format(next(iter(apps.keys()))))
```

src/preload.py
```diff
@@ -749,6 +758,7 @@ def main_preload(app_data, additional_bytes):
     if flasher_root:
         flasher_root.resize(additional_bytes)
         with flasher_root.mount_context_manager() as mountpoint:
+            write_resin_device_pinning(app_data, mountpoint + "/etc/resin-device-pinning.conf")
```
zvin: This line is too long, please run flake8 on the python code.
```python
        write_resin_device_pinning(
            app_data,
            mountpoint + "/etc/resin-device-pinning.conf"
        )
```

PR #183, opened 2018-08-27
windows: Running the executable and flushing stdout

imrehg:
The preloader container is built on the user's system. On Windows, the files don't retain their executable flags, thus the preloader Python script can't be started just by running the script. Now explicitly running the script with `python3`, so whether or not the script is executable, it will run.

When attaching to a docker container, docker will buffer the standard input/output. On Windows the script seemed to hang as the reply from the preloader container might be quite small, thus docker just keeps waiting for more data before giving it to the program that attached to it. Flushing stdout seems to have the desired effect and the caller receives the reply from the preloader.

Change-type: patch
Signed-off-by: Gergely Imreh <gergely@resin.io>

Front logo [Front conversations](https://app.frontapp.com/open/top_dg3n)


brownjohnf:
@resin-ci retest

PR #184, opened 2018-08-29
Use localhost instead of 0.0.0.0 to connect to Docker on win32

zvin:
Change-type: patch
Signed-off-by: Alexis Svinartchouk <alexis@resin.io>

Front logo [Front conversations](https://app.frontapp.com/open/top_dhbn)

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -682,9 +682,11 @@ class Preloader extends EventEmitter {
     .then((registryToken) => {
       const opts = { authconfig: { registrytoken: registryToken } }
       // Docker connection
+      // We use localhost on windows because of this bug in node < 8.10.0:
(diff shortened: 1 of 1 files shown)


lib/preload.js
```diff
@@ -682,9 +682,11 @@ class Preloader extends EventEmitter {
     .then((registryToken) => {
       const opts = { authconfig: { registrytoken: registryToken } }
       // Docker connection
+      // We use localhost on windows because of this bug in old node versions:
```
imrehg: Would probably spell out that node version <8.10.0, but that's just cherry on the top :)

imrehg:
Tested on previously breaking v8.9.4 and v6.14.4 node versions, and with this patch preloading works.

zvin:
@resin-ci retest

Issue #194, opened 2018-12-28
Preload does not work with openBalena instances using self-signed certificates


cjgordon:
When trying to preload and image using openBalena instance with self signed certificates I get the following error:
    
    $ balena preload balena-intel-nuc-2.29.0-v9.0.1.img --app myApp
    Building Docker preloader image. [========================] 100%
    / Creating preloader container
    | Starting preloader container
    / Fetching application gateway
    / Reading image information
    ? Select a release latest
    \ Resizing partitions and waiting for dockerd to start
    - Cleaning up temporary files
    (HTTP code 500) server error - Get https://registry.myDomain.com/v2/: x509: certificate signed by 
    unknown authority 

I believe the instance of docker daemon being spun up to preload the container images onto the image does not have the correct CA certificate configuration to trust the self-signed certificate. I have tried using the --ca and --key options but I believe these are for authenticating with docker host specified with --dockerHost not the image registry.

Cheers
Chris

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_gfgr)

zvin:
This should be fixed on Linux hosts in `v8.1.0` with https://github.com/balena-io/balena-preload/pull/195 .
Not sure how to implement this on other OSes for now.

dfunckt:
Regarding cross-platfrom support — quoting you from that PR:

> As for passing certificates explicitly, I've found this https://docs.docker.com/engine/security/certificates/ . Implementation will be difficult. If we ask the user for a .pem, we need to extract the .crt and the domain name from it and place it into /etc/docker/certs.d/DOMAIN/ca.crt in the container.

In a later step we could potentially use openssh to find the domain name automatically but for now requiring the user to explicitly specify the cert and domain name sounds reasonable to me.

zvin:
Use the `--add-certificate` option.

PR #195, opened 2019-01-07
Bind mount host's /etc/ssl/certs folder on Linux

zvin:
Connects-to: #194 
Change-type: minor

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -133,6 +133,9 @@ const createContainer = (docker, image, splashImage, dockerPort, proxy, edisonFo
       // In some situations, devices created by `losetup -f` in the container only appear on the host and not in the container.
       // See https://github.com/balena-io/balena-cli/issues/1008
       mounts.push(bindMount('/dev', '/dev', version.ApiVersion))
+      // Use host's ssl certificates
+      // TODO: how can we achieve this on Windows and macOS?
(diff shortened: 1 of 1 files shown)


zvin:
@dfunckt 
The preloader container should trust the certificates from the host os. As a user I would expect this behaviour.
This isn't cross-platform, I'd like to find a way to make it so. This PR is a quick fix for the user that encountered the problem.
Adding the host's certificates in the preloader container is more complicated on Windows and macOS.

As for passing certificates explicitly, I've found this https://docs.docker.com/engine/security/certificates/ .
Implementation will be difficult. If we ask the user for a `.pem`, we need to extract the `.crt` and the domain name from it and place it into `/etc/docker/certs.d/DOMAIN/ca.crt`.

dfunckt:
I see. I think though that Docker simply ignores the system cert store and just uses its own so the mount like you do here won't work, isn't that right?

dfunckt:
I may be misremembering though and what I say only apply to Docker for Mac and the VM it launches in the background.

zvin:
@dfunckt I just checked bind mounting an empty folder at `/etc/ssl/certs` in the preloader container. I get this error:
`(HTTP code 500) server error - Get https://registry2.balena-cloud.com/v2/: x509: failed to load system roots and no roots provided`
which means that `dockerd` uses system certs.

dfunckt:
Perfect, thanks for confirming and also coming up with a fix so fast in the first place @zvin !

PR #201, opened 2019-05-09
Filter by commit as part of the query if it's specified

Page-:
Change-type: patch

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -408,7 +408,7 @@ class Preloader extends EventEmitter {
 
   _getRelease () {
     const commit = this._getCommit()
-    const releases = _.orderBy(this.application.owns__release, [ 'id' ], [ 'desc' ])
+    const releases = this.application.owns__release
     if ((commit === null) && releases.length) {
       return releases[0]
     }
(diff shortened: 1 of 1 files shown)


lib/preload.js
```diff
@@ -494,7 +494,31 @@ class Preloader extends EventEmitter {
...
+                  image: {
+                    $select: ['image_size', 'is_stored_at__image_location']
+                  }
+                }
+              }
+            },
+            $filter: releaseFilter
+          }
+        }
+      })
```
zvin: @Page- can't we reuse `preload.applicationExpandOptions` here?

Something like
```javascript
const options = _.cloneDeep(preload.applicationExpandOptions);
if (this.commit != null) {
    options.owns__release.$filter.commit = { $startswith: this.commit }
}
return this.balena.models.application.get(this.appId, {$expand: options})
```

Page-: So the plan was to move more of the querying/filtering into this directly, but because of the way all the usage is spread out I haven't managed to move more filters in by eye - but this one alone gives us a workaround for the issue for now at least

PR #202, opened 2019-05-15
When checking supervisor version strip logstream ending too

imrehg:
Some supervisors have `vX.Y.Z` and some `vX.Y.Z_logstream` entries, and all should fall back in the version check to the basic `X.Y.Z` values.

Change-type: patch
Signed-off-by: Gergely Imreh <gergely@balena.io>


imrehg:
Superceded by #203

PR #209, opened 2019-07-18
Fix certificates path in the preloader container

zvin:
Change-type: patch


zvin:
A user reported that docker was not using the added certificate: https://forums.balena.io/t/open-balnea-preload-image-error-x509-certificate-signed-by-unknown-authority/4824/64
Somehow it was working for `curl` / `wget` but not for `dockerd`.

PR #220, opened 2020-07-07
Increase losetup retry delay and count, and print full stderr for dockerd

pdcastro:
Connects-to: balena-io/balena-cli/issues/1922
Change-type: minor

### DIFF
--- a/src/preload.py
+++ b/src/preload.py
@@ -65,6 +65,52 @@ log.addHandler(StreamHandler())
 PARTITIONS_CACHE = {}
 
 
+class RetryCounter:
+    """Counter and logger for the number of times that a function is retried.
+    Usage:
+        retry_counter = RetryCounter()
+        hint = "Try reducing the system entropy"
+        wrapped_func, wrap_key = retry_counter.wrap(my_func, hint, *args)
+        retry_call(wrapped_func, fargs=args, ...)
+        retry_counter.clear(wrap_key)
+    """
+
+    def __init__(self):
+        self.counter = {}
+
+    @staticmethod
+    def key(func_name, *args):
+        return " ".join(str(e) for e in (func_name,) + args)
+
+    def clear(self, key, *args):
+        del self.counter[key]
+
+    def inc(self, key, *args):
+        self.counter[key] = self.counter.setdefault(key, 0) + 1
+        return self.counter[key]
+
+    def wrap(self, func, hint, *args, **kwargs):
+        """Return a function that wraps the given func, counting its usage"""
+        key = self.key(func.__name__, *args)
+
+        def wrapped(*args, **kwargs):
+            count = self.inc(key, *args)
+            if count > 1:
+                log.info(
+                    "\nRetrying (count={}) {}\n{}".format(
+                        count,
+                        key,
+                        hint,
+                    ),
+                )
+            return func(*args, **kwargs)
+
+        return (wrapped, key)
+
+
+retry_counter = RetryCounter()
+
+
 def get_partitions(image):
     return {p.label: p for p in PartitionTable(image).partitions if p.label}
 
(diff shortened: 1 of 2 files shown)


src/preload.py
```diff
@@ -79,6 +79,33 @@ def prepare_global_partitions():
...
+        return self.counter[key]
+
+retry_counter = RetryCounter()
+
+def losetup_wrap(*args, **kwargs):
+    count = retry_counter.inc(*args)
+    if count > 1:
+        msg = """
+Retrying (count={}) losetup {}
+Hint: if using a Virtual Machine, consider increasing the CPU count.
```
pdcastro, line 105: I've learned this hint the hard way, :-) using a VirtualBox VM with Debian, on a macOS host. With the default CPU count of 1, `losetup` would fail numerous times, even though the the overall laptop resource usage was relatively low (free physical memory, VM not swapping, VM CPU between 50% and 100%). Increasing the CPU count to 2 reduces the number of losetup retries a lot, and increasing it to 4 makes losetup succeed without any retries (almost always, occasionally a single retry).

src/preload.py
```diff
@@ -79,6 +79,33 @@ def prepare_global_partitions():
...
+
+    def clear(self, *args):
+        del self.counter[self.key(*args)]
+
+    def inc(self, *args):
+        key = self.key(*args)
+        self.counter[key] = self.counter.setdefault(key, 0) + 1
+        return self.counter[key]
+
+retry_counter = RetryCounter()
```
zvin, line 98: No need for a counter and a wrapper:

```python3
class LosetupLogger:
    @staticmethod
    def warning(fmt, error, delay):
        msg = """
Retrying losetup in {} seconds
Error: {}
Hint: if using a Virtual Machine, consider increasing the CPU count.
""".format(delay, error)
        log.info(msg)


@contextmanager
def losetup_context_manager(image, offset=None, size=None):
    args = ["-f", "--show"]
    if offset is not None:
        args.extend(["--offset", offset])
    if size is not None:
        args.extend(["--sizelimit", size])
    args.append(image)
    # In the case of slow hardware the kernel might be in the middle of
    # tearing down internal structure
    device = retry_call(
        losetup,
        fargs=args,
        tries=10,
        delay=3,
        max_delay=30,
        backoff=2,
        logger=LosetupLogger(),
    ).stdout.decode("utf8").strip()
    yield device
    losetup("-d", device)
```

src/preload.py
```diff
@@ -65,6 +65,47 @@
...
+
+    def wrap(self, func, hint, *args, **kwargs):
+        """Return a function that wraps the given func, counting its usage"""
+        key = self.key(*args)
+
+        def wrapped(*args, **kwargs):
+            count = self.inc(key, *args)
+            if count > 1:
+                log.info("\nRetrying (count={}) {} {}\n{}".format(
+                    count, func.__name__, key, hint))
```
zvin, line 100:
```suggestion
                log.info(
                    "\nRetrying (count={}) {} {}\n{}".format(
                        count,
                        func.__name__,
                        key,
                        hint,
                    ),
                )
```

pdcastro: Done, plus one other small change to the `RetryCounter.key()` method: the function name was added to the `key` computation (previously only the function arguments were included, not the function name).

src/preload.py
```diff
@@ -65,6 +65,52 @@
...
+        self.counter = {}
+
+    @staticmethod
+    def key(func_name, *args):
+        return " ".join(str(e) for e in (func_name,) + args)
+
+    def clear(self, key, *args):
+        del self.counter[key]
+
+    def inc(self, key, *args):
```
zvin, line 88: I've approved a bit too fast.
`*args` are not used here.

zvin: nor in `clear`

pdcastro: Indeed. Harmless though (tested). I'll be testing further during integration with the CLI, and then I'll remove the unused args in another PR. (Waiting in case additional changes are also needed.)

PR #226, opened 2020-08-04
Update balena-sdk to 15.x

Page-:
Update balena-sdk from 14.0.0 to 15.2.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #230, opened 2020-08-24
Fix splash image file name for balenaOS >= 2.53.0 (resin → balena)

pdcastro:
Connects-to: balena-io/balena-cli/issues/1993

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -20,7 +20,7 @@ const { R_OK, W_OK } = fs.constants;
 
 const DOCKER_IMAGE_TAG = 'balena/balena-preload';
 const DISK_IMAGE_PATH_IN_DOCKER = '/img/balena.img';
-const SPLASH_IMAGE_PATH_IN_DOCKER = '/img/resin-logo.png';
+const SPLASH_IMAGE_PATH_IN_DOCKER = '/img/balena-logo.png';
 const DOCKER_STEP_RE = /Step (\d+)\/(\d+)/;
 const EXIT_CAUSED_BY_SIGINT = 137;
 
@@ -240,6 +240,7 @@ class Preloader extends EventEmitter {
 		this.freeSpace = null; // space available on the image data partition (in bytes)
 		this.preloadedBuilds = null; // list of preloaded Docker images in the disk image
 		this.supervisorVersion = null; // disk image supervisor version
+		this.balenaOSVersion = null; // OS version from the image's "/etc/issue" file
 		this.config = null; // config.json data from the disk image
 		this.deviceTypes = null;
 	}
@@ -427,6 +428,7 @@ class Preloader extends EventEmitter {
 				this.freeSpace = info.free_space;
 				this.preloadedBuilds = info.preloaded_builds;
 				this.supervisorVersion = info.supervisor_version;
+				this.balenaOSVersion = info.balena_os_version;
 				this.config = info.config;
 			});
 		});
@@ -803,6 +805,21 @@ class Preloader extends EventEmitter {
 		}
 	}
 
+	/**
+	 * Return the splash image path depending on the balenaOS version.
+	 * (It was renamed from `resin-logo` to `balena-logo` in balenaOS v2.53.0.)
+	 */
+	_getSplashImagePath() {
+		try {
+			if (compareVersions(this.balenaOSVersion, '2.53.0') >= 0) {
+				return '/splash/balena-logo.png';
+			}
+		} catch (err) {
+			// invalid semver (including the empty string, undefined or null)
+		}
+		return '/splash/resin-logo.png';
+	}
+
 	preload() {
 		let images;
 		return this._getState()
@@ -820,6 +837,7 @@ class Preloader extends EventEmitter {
 						return this._runCommand('preload', {
 							app_data: this._getAppData(),
 							additional_bytes: additionalBytes,
+							splash_image_path: this._getSplashImagePath(),
 						});
 					},
 				);
--- a/src/preload.py
+++ b/src/preload.py
@@ -42,8 +42,7 @@ MBR_SIZE = 512
 GPT_SIZE = SECTOR_SIZE * 34
 MBR_BOOTSTRAP_CODE_SIZE = 446
 
-SPLASH_IMAGE_FROM = "/img/resin-logo.png"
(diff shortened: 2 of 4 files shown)


lib/preload.js
```diff
@@ -20,7 +20,7 @@ const { R_OK, W_OK } = fs.constants;
 
 const DOCKER_IMAGE_TAG = 'balena/balena-preload';
 const DISK_IMAGE_PATH_IN_DOCKER = '/img/balena.img';
-const SPLASH_IMAGE_PATH_IN_DOCKER = '/img/resin-logo.png';
+const SPLASH_IMAGE_PATH_IN_DOCKER = '/img/balena-logo.png';
```
pdcastro, line 23: This name only applies to the Docker container created by balena-preload in order to execute the `preload.py` Python program. It is not the name of the splash image file in the preloaded balenaOS image.

lib/preload.js
```diff
@@ -427,6 +428,7 @@ class Preloader extends EventEmitter {
 				this.freeSpace = info.free_space;
 				this.preloadedBuilds = info.preloaded_builds;
 				this.supervisorVersion = info.supervisor_version;
+				this.balenaOSVersion = info.balenaOS_version;
```
pdcastro, line 431: This version is extracted from the `/etc/issue` file in the preloaded balenaOS image.

src/preload.py
```diff
@@ -42,8 +43,9 @@
 GPT_SIZE = SECTOR_SIZE * 34
 MBR_BOOTSTRAP_CODE_SIZE = 446
 
-SPLASH_IMAGE_FROM = "/img/resin-logo.png"
-SPLASH_IMAGE_TO = "/splash/resin-logo.png"
+SPLASH_IMAGE_FROM = "/img/balena-logo.png"
+SPLASH_IMAGE_TO = "/splash/balena-logo.png"
+OLD_SPLASH_IMAGE_TO = "/splash/resin-logo.png"
```
pdcastro, line 48: These are the possible names of the splash image file in the preloaded balenaOS image.

src/preload.py
```diff
@@ -715,8 +729,18 @@ def get_inner_image_path(root_mountpoint):
...
+    of the "/etc/issue" file in the etc_issue_contents argument.
+    """
+    m = match('balenaOS (.+?) ', etc_issue_contents)
+    return m[1] if m is not None else ""
+
+
 def _get_images_and_supervisor_version(image=None):
-    driver = get_docker_storage_driver(image=image)
+    rootA_file_contents = get_rootA_file_contents(image)
+    driver = get_docker_storage_driver(rootA_file_contents.get("docker_service", ""))
```
pdcastro, line 743: This is part of an optimization to avoid mounting the rootA partition twice (once for reading `/lib/systemd/system/balena.service` and then again for reading `/etc/issue`). The partition is mounted only once, and contents of those files saved to the `rootA_file_contents` object.

src/preload.py
```diff
@@ -814,18 +842,30 @@ def get_docker_service_file_path(folder):
             return fpath
 
 
-def get_docker_service_file_content(image=None):
+def get_rootA_file_contents(image=None):
+    file_contents = {
+        "docker_service": "",
+        "etc_issue": "",
+    }
```
pdcastro, line 849: This is an optimization to avoid mounting the rootA partition twice (once for reading `/lib/systemd/system/balena.service` and then again for reading `/etc/issue`). The partition is mounted only once, and contents of those files saved to the `file_contents` object which is returned.

src/preload.py
```diff
@@ -836,8 +876,8 @@ def find_one_of(lst, *args):
     return -1
 
 
-def get_docker_storage_driver(image=None):
-    for line in get_docker_service_file_content(image).strip().split("\n"):
+def get_docker_storage_driver(docker_service_file_contents):
+    for line in docker_service_file_contents.strip().split("\n"):
```
pdcastro, line 880: This is part of an optimization to avoid mounting the rootA partition twice (once for reading `/lib/systemd/system/balena.service` and then again for reading `/etc/issue`). The partition is mounted only once, and contents of those files saved to the `rootA_file_contents` object. The contents of the `balena.service` file are given as the input to this function.

src/preload.py
```diff
@@ -613,18 +615,29 @@ def write_apps_json(data, output):
         json.dump(data, f, indent=4, sort_keys=True)
 
 
-def replace_splash_image(image=None):
+def replace_splash_image(balenaOS_version="", image=None):
```
zvin, line 618: NIT: `balena_os_version` would look better.

There is no need for a default value, get_baleaOS_version always returns a string.

src/preload.py
```diff
@@ -715,8 +729,18 @@ def get_inner_image_path(root_mountpoint):
         )
 
 
+def _get_balenaOS_version(etc_issue_contents):
```
zvin, line 732: NIT: `get_balena_os_version` would look better

src/preload.py
```diff
@@ -814,18 +842,30 @@ def get_docker_service_file_path(folder):
...
-        with open(path) as f:
-            return f.read()
+        root_folder = docker_root if docker_root is not None else mountpoint
+        docker_service_path = get_docker_service_file_path(root_folder) or ""
+        etc_issue_path = os.path.join(root_folder, "etc", "issue")
+        with open(docker_service_path) as f:
+            file_contents["docker_service"] = f.read()
+        try:
+            with open(etc_issue_path) as f:
+                file_contents["etc_issue"] = f.read()
```
zvin, line 862: NIT: the key could be `/etc/issue` or just `issue`

src/preload.py
```diff
@@ -846,7 +886,7 @@ def get_docker_storage_driver(image=None):
     assert False, "Docker storage driver could not be found"
 
 
-def main_preload(app_data, additional_bytes):
+def main_preload(app_data, additional_bytes, balenaOS_version=""):
```
zvin, line 889: There is no need for a default value.

zvin: Instead of passing the version here, you could pass the splash image destination (`SPLASH_IMAGE_TO`), that way you could use the already installed `semver` on the node side and not have to add it in `requirements.txt`.

pdcastro:
Thanks for the review, @zvin 👍   I believe I have addressed all of your suggestions.

zvin (approved):

Issue #233, opened 2020-09-24
Error handling: python `sh` output is truncated, and uncaught exceptions are not converted to JSON


pdcastro:
Currently, the output of commands (such as `dockerd`) executed with the python 'sh' module is truncated, for example:
```
  RAN: /usr/local/bin/dockerd --storage-driver=aufs --data-root=/tmp/tmpay62r387/docker --host=tcp://0.0.0.0:56360

  STDOUT:

  STDERR:
time="2020-09-23T19:52:59.762672297Z" level=warning msg="[!] DON'T BIND ON ANY IP ADDRESS WITHOUT setting --tlsverify IF YOU DON'T KNOW WHAT YOU'RE DOING [!]"
time="2020-09-23T19:52:59.764371255Z" level=info msg="libcontainerd: started new docker-containerd process" pid=233
time="2020-09-23T19:52:59Z" level=info msg="starting containerd" module=containerd revision=89623f28b87a6004d4b785663257362d1658a729 version=v1.0.0
time="2020-09-23T19:52:59Z" level=info msg="setting subreaper..." module=containerd
time="2020-09-23T19:52:59Z" level=info msg="changing OOM score to -500" module=containerd
time="2020-09-23T19:52:59Z" level=info msg="loading plugin "io.containerd.content.v1.content"..." module=containerd type=io.containerd.content.v1
tim... (3787 more, please see e.stderr)
```

Also, the interprocess communication between `preload.js` and `preload.py` exchanges JSON objects over stdin / stdout, but unexpected exceptions are not always captured and converted to JSON.

pdcastro:
Connects-to: balena-io/balena-cli/issues/2045

pdcastro:
This issue is resolved by a combination of PRs #239, #240 and #241.

PR #234, opened 2020-09-24
Don't truncate error logs, improve error handling 

pdcastro:
Resolves: #233 
Connects-to: balena-io/balena-cli/issues/2045

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -403,26 +403,21 @@ class Preloader extends EventEmitter {
 				try {
 					result = JSON.parse(data).result;
 				} catch (error) {
-					result.StatusCode = 1;
-					result.Error = error;
+					result.error = error;
 				}
-				if (result.StatusCode === 0) {
-					resolve(result);
-				} else {
+				if (result.error) {
 					const msg = [
 						`An error has occurred executing internal preload command '${command}':`,
 						cmd,
 					];
-					if (result.Error) {
-						msg.push(
-							`StatusCode: ${result.StatusCode}`,
-							`Error: ${result.Error}`,
-						);
-					} else {
-						msg.push('Result:', JSON.stringify(result, null, 4));
-					}
+					msg.push(
+						`StatusCode: 1`,
+						`Error: ${result.error}`,
+					);
 					msg.push('');
 					reject(new Error(msg.join('\n')));
+				} else {
+					resolve(result);
 				}
 			});
 			this.stdin.write(cmd);
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -24,7 +24,6 @@ const DOCKER_IMAGE_TAG = 'balena/balena-preload';
 const DISK_IMAGE_PATH_IN_DOCKER = '/img/balena.img';
 const SPLASH_IMAGE_PATH_IN_DOCKER = '/img/balena-logo.png';
 const DOCKER_STEP_RE = /Step (\d+)\/(\d+)/;
-const EXIT_CAUSED_BY_SIGINT = 137;
 const CONCURRENT_REQUESTS_TO_REGISTRY = 10;
 
 const GRAPHDRIVER_ERROR =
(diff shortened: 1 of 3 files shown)


lib/preload.js
```diff
@@ -352,32 +351,72 @@ class Preloader extends EventEmitter {
...
+	 *
+	 * @param {string} command
+	 * @param {string[]} parameters
+	 */
 	_runCommand(command, parameters) {
 		return new Bluebird((resolve, reject) => {
+			const cmd = JSON.stringify({ command, parameters }) + '\n';
 			this.stdout.once('error', reject);
 			this.stdout.once('data', (data) => {
-				resolve(JSON.parse(data).result);
```
pdcastro, line 378: As I understood it... This `_runCommand` function is one "end" (the CLI process end) of the interprocess communication between:
* The balena CLI / `preload.js` process at one end, and
* `preload.py` running in a Docker container at the other end.

The interprocess communication uses JSON objects exchanged via stdin and stdout. However, in case `preload.py` failed for any reason, a JSON object might not be generated in its stdout, and this `JSON.parse(data).result` call could fail synchronously (rather than producing a rejected promise).

In addition, the interprocess communication has to concept of "commands" that are sent via stdin to be executed in a Docker container, producing the result in stdout. If the command produced a JSON object with StatusCode different than zero (i.e. a failed execution), at least some commands like `get_image_info` would ignore the error: `_runCommand` would resolve rather than reject, and the resolved object's `StatusCode` would also not be inspected anywhere else (like elsewhere in `preload.js`, or elsewhere in `balena-cli`).

I though it would make sense for `_runCommand` to inspect `StatusCode` and produce a rejected promise with an error if `StatusCode` was not zero. This is what my changes to `_runCommand` implement.

package.json
```diff
@@ -59,6 +59,7 @@
   "scripts": {
     "lint": "balena-lint -e js --typescript lib",
     "lint-fix": "balena-lint -e js --typescript --fix lib",
+    "lint-python": "flake8 src/preload.py",
```
pdcastro, line 62: This npm script was added as "hint", for someone to execute it manually if they know and remember about it. Alexis mentioned it to me in this older PR: https://github.com/balena-io-modules/balena-preload/pull/220#pullrequestreview-446302471

Automated execution is a "To Do". The `flake8` command is a python utility, not `npm` package, so cannot (?) be installed with something like "npm install", and adding some sort of "pip install" would mean having to deal with setting up a python environment or Dockerfile... Hmm, actually, there is already a Dockerfile and we could add some `docker run` command... Well, "to do", I didn't mean to solve this issue in this PR, just leave this npm script as a hint. :-)

src/preload.py
```diff
@@ -53,14 +53,17 @@
 
 SUPERVISOR_REPOSITORY_RE = "^resin(playground)?/[a-z0-9]+-supervisor$"
 
+# 'sh' module '_truncate_exc' option:
+# https://github.com/amoffat/sh/blob/1.12.14/sh.py#L1151
+SH_OPTS = {"_truncate_exc": False}
```
pdcastro, line 58: This `{"_truncate_exc": False}` option for the Python `sh` module prevents the "friendly" truncation of error output.

src/preload.py
```diff
@@ -78,22 +81,24 @@ def __init__(self):
...
     @staticmethod
-    def key(func_name, *args):
-        return " ".join(str(e) for e in (func_name,) + args)
+    def key(func_name, *args, **kwargs):
+        return " ".join(
+            str(e) for e in (func_name,) + args + tuple(kwargs.values())
+        )
 
-    def clear(self, key, *args):
+    def clear(self, key):
```
pdcastro, line 89: Fix of a harmless bug, finally addressing Alexis' request in this older PR comment: https://github.com/balena-io-modules/balena-preload/pull/220#discussion_r455057117

src/preload.py
```diff
@@ -538,16 +548,8 @@ def start_docker_daemon(storage_driver, docker_dir):
...
-                running_dockerd.wait()
-            except ErrorReturnCode as e:
-                log.info(
-                    "An error has occurred executing 'dockerd':\n{}".format(
-                        e.stderr.decode("utf8"),
-                    ),
-                )
-                raise e
+            # This will raise an sh.ErrorReturnCode_X exception.
+            running_dockerd.wait()
```
pdcastro, line 552: This change reverts to how it was before I changed it in the previous PR: https://github.com/balena-io-modules/balena-preload/pull/220/files#diff-5d1069846576354d1dab375d260d34c1L493

There was nothing wrong the the code, but the `log.info()` proved redundant after this PR introduced the `{"_truncate_exc": False}` to `sh`. (The `log.info()` was a workaround to show the full error output when `DEBUG=1` was set by the CLI.)

src/preload.py
```diff
@@ -947,10 +952,16 @@ def get_partition(name, image=None):
...
         data = json.loads(line)
         method = methods[data["command"]]
-        result = method(**data.get("parameters", {}))
+        try:
+            result = method(**data.get("parameters", {}))
+        except BaseException as e:
+            result = {
+                "StatusCode": 1,
+                "Error": str(e),
+            }
```
pdcastro, line 965: This `try/except` block aims at otherwise uncaught exceptions, attempting to ensure that a valid JSON object is always produced in the stdout of `preload.py`, even in the event of unexpected errors. The `_runCommand` function in `preload.js` was amended accordingly.

src/preload.py
```diff
@@ -956,6 +956,12 @@ def get_partition(name, image=None):
     for line in sys.stdin:
         data = json.loads(line)
         method = methods[data["command"]]
-        result = method(**data.get("parameters", {}))
+        try:
+            result = method(**data.get("parameters", {}))
+        except BaseException as e:
```
zvin, line 961: `except Exception as e` should be fine

src/preload.py
```diff
@@ -956,6 +956,12 @@ def get_partition(name, image=None):
...
+            result = {
+                "StatusCode": 1,
```
zvin, line 963: I don't really get the point of `StatusCode`.
It will never be anything else than `1`.

`StatusCode` comes from dockerode's container.wait(), we shouldn't reuse this name

src/preload.py
```diff
@@ -956,6 +956,12 @@ def get_partition(name, image=None):
...
+                "Error": str(e),
```
zvin, line 964: Why start with capital letters ?

lib/preload.js
```diff
@@ -352,32 +351,72 @@ class Preloader extends EventEmitter {
...
+				let result = {};
+				try {
+					result = JSON.parse(data).result;
+				} catch (error) {
+					result.StatusCode = 1;
+					result.Error = error;
+				}
+				if (result.StatusCode === 0) {
```
zvin, line 400: This will never happen.

lib/preload.js
```diff
@@ -361,32 +360,72 @@ class Preloader extends EventEmitter {
...
+	 */
 	_runCommand(command, parameters) {
 		return new Bluebird((resolve, reject) => {
+			const cmd = JSON.stringify({ command, parameters }) + '\n';
 			this.stdout.once('error', reject);
 			this.stdout.once('data', (data) => {
-				resolve(JSON.parse(data).result);
+				let result = {};
+				try {
+					result = JSON.parse(data).result;
```
klutchell, line 404:
```suggestion
					result = JSON.parse(data).result;
```
```suggestion
					result = JSON.parse(data);
```
I think by specifying `JSON.parse(...).result` you will assign a boolean to `result`, and not the parsed data. This would explain Alexis' comment below. I've made the change in my workspace for now.

klutchell, line 404:
```suggestion
					result = JSON.parse(data);
```

src/preload.py
klutchell:
```suggestion
        except Exception as e:
```

pdcastro:
There were 3 commits -- I've taken the 2 uncontroversial ones out to PR #239 (self certified and merged), so that this PR can focus on the 3rd one.

lib/preload.js
```diff
@@ -361,32 +360,67 @@ class Preloader extends EventEmitter {
...
+					result.error = error;
+				}
+				if (result.error) {
+					const msg = [
+						`An error has occurred executing internal preload command '${command}':`,
+						cmd,
+					];
+					msg.push(
+						`StatusCode: 1`,
+						`Error: ${result.error}`,
```
klutchell, line 415: Here we still want to return `StatusCode` and `Error` for any calling functions that are expecting for dockerode errors in that format.

lib/preload.js
```diff
@@ -361,32 +360,67 @@ class Preloader extends EventEmitter {
...
+	 * have failed and the returned promise is rejected with an error message
+	 * including the message provided in the `Error` field (if any).
+	 *
+	 * @param {string} command
+	 * @param {object} parameters
+	 */
 	_runCommand(command, parameters) {
 		return new Bluebird((resolve, reject) => {
+			const cmd = JSON.stringify({ command, parameters }) + '\n';
 			this.stdout.once('error', reject);
```
klutchell, line 400: A normal dockerode failure will fall into here with `StatusCode` and `Error` having some values. Those values will never be present in the `data` callback.

lib/preload.js
pdcastro: Thanks your comments @klutchell 👍 
I found a few other issues while testing the code -- it was properly broken, not sure how I missed all of it before creating this PR last year. 😬
* I have moved some more code out of this PR to further focus the review into the "interprocess communication" aspect. Besides PR #239, some code was moved to PR #241, also self certified and merged. Feel free to review it post mortem if you have any concerns!
* For the remaining code, I didn't want to overwrite your fixup commit here so I created PR #240, which is looking more promising in my testing. Could you have a look at that PR, and move any comments there if you think they still apply? 🙏

lib/preload.js
pdcastro: My main concern in ensure that the CLI process terminates (no infinite spinning wheels), preferably with a helpful error message. If `reject` is called here, I think the rejection bubbles up to the CLI and the process terminates gracefully.

Are you suspecting that the error message would not be helpful?

pdcastro:
This PR was closed in favour of #239, #240 and #241.

PR #237, opened 2021-04-26
Bump Docker version to 20.10.6

klutchell:
Docker 20.10 supports cgroup v2 and works on Fedora by default.

Change-type: patch
Changelog-entry: Bump Docker version to 20.10.6

See these posts relating to the support for cgroups v2 in Docker:
- https://www.docker.com/blog/introducing-docker-engine-20-10/
- https://medium.com/nttlabs/cgroup-v2-596d035be4d7

Resolves:
- https://github.com/balena-io/balena-cli/issues/2250

I first reproduced the issue with a Fedora 33 VM.
- installing Docker v19 would not even run `hello-world`
- installing Docker v20 was able to run `hello-world`
- running `balena-preload` would fail to start the Docker-in-Docker container
- bumping the dind version from v17 to v20 allowed Docker-in-Docker to run

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM docker:17.12.0-ce-dind
+FROM docker:20.10.6-dind
 
 # coreutils so we have the real dd, not the busybox one
-RUN apk update && apk add --no-cache python3 parted btrfs-progs util-linux sfdisk file coreutils sgdisk
+RUN apk update && apk add --no-cache py3-pip parted btrfs-progs util-linux sfdisk file coreutils sgdisk
 
 COPY ./requirements.txt /tmp/
 


Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM docker:17.12.0-ce-dind
+FROM docker:20.10.6-dind
```
pdcastro, line 1: Nice 👍   Does this also work when the host OS / VM has support for cgroups v1 only, and is running an older version of Docker like Docker v18? I understand that Ubuntu v18.04 supports both cgroups v1 and v2, as the following command works:
```
sudo mount -t cgroup2 none /mnt/test
```
So it would probably have to be an older version of Ubuntu or some other distro, where cgroups v2 is not available.

I ask because, ideally, `balena preload` would work on both older and newer distros.

klutchell: You're right, I assumed that docker v20 has support for both cgroups v1 and v2 but I failed to test this. I'll see if I can find a distro that only supports v1 and if Docker 20 can run `hello-world` then I think we have our answer.

klutchell: Actually cgroups v2 has been in the kernel since 2014, so we really only need to test distros where v1 is the default and where v2 is the default.

> Fedora 31 was released on October 29, 2019. This is the first major distro that comes with cgroup v2 (aka unified hierarchy) enabled by default, 5 years after it first appeared in Linux kernel 3.16 (Aug 3, 2014).

klutchell: I confirmed this works on a `ubuntu/bionic64` VM image as well which is using cgroups v1 by default.

pdcastro (approved):

PR #238, opened 2021-05-03
Don't assume that 'docker' argument uses Bluebird promises

pdcastro:
Change-type: patch


lib/preload.js
```diff
@@ -1075,11 +1076,13 @@ class Preloader extends EventEmitter {
...
 				// We use localhost on windows because of this bug in node < 8.10.0:
 				// https://github.com/nodejs/node/issues/14900
-				const innerDockerProgress = new dockerProgress.DockerProgress({
-					Promise: Bluebird,
+				const innerDocker = new Docker({
 					host: os.platform() === 'win32' ? 'localhost' : '0.0.0.0',
 					port: this.dockerPort,
 				});
+				const innerDockerProgress = new dockerProgress.DockerProgress({
+					docker: innerDocker,
```
pdcastro, line 1084: This is done to use the new interface of `docker-progress` v5:
https://github.com/balena-io-modules/docker-progress/pull/54

It allows `docker-progress` to use the same version (and even the same instance) of `dockerode` as `balena-preload` and `balena-cli`.

package.json
```diff
@@ -23,8 +23,8 @@
     "balena-sdk": "^15.3.1",
     "bluebird": "^3.7.2",
     "compare-versions": "^3.6.0",
-    "docker-progress": "^4.0.3",
-    "dockerode": "^2.5.8",
+    "docker-progress": "^5.0.0",
+    "dockerode": "^3.3.0",
```
pdcastro, line 27: The plan is to get the CLI to use `dockerode` v3 as well:
https://github.com/balena-io/balena-cli/pull/2259/files

pdcastro:
@balena-ci I self-certify!

PR #240, opened 2021-05-04
Enhance comms between CLI process and Python process in container

pdcastro:
Connects-to: #233 
Connects-to: balena-io/balena-cli/issues/2045

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -382,13 +382,56 @@ class Preloader extends EventEmitter {
 			.catch((error) => this.emit('error', error));
 	}
 
+	/**
+	 * Send a command to `preload.py` (running in a Docker container) through
+	 * its stdin, and read a response from its stdout. The stdout response
+	 * must be a JSON object containing a `statusCode` field (integer) and
+	 * optionally `result` and `error` fields. If `statusCode` is the number
+	 * zero, the command execution is considered successful, otherwise it is
+	 * assumed to have failed and the returned promise is rejected with an
+	 * error message including the message provided in the `error` field.
+	 *
+	 * @param {string} command
+	 * @param {object} parameters
+	 */
 	_runCommand(command, parameters) {
 		return new Bluebird((resolve, reject) => {
+			const cmd = JSON.stringify({ command, parameters }) + '\n';
 			this.stdout.once('error', reject);
 			this.stdout.once('data', (data) => {
-				resolve(JSON.parse(data).result);
+				let strData = data;
+				try {
+					strData = data.toString();
+				} catch (_e) {
+					// ignore
+				}
+				let response = {};
+				try {
+					response = JSON.parse(strData);
+				} catch (error) {
+					response.statusCode = 1;
+					response.error = error;
+				}
+				if (response.statusCode === 0) {
+					resolve(response.result);
+				} else {
+					const msg = [
+						`An error has occurred executing internal preload command '${command}':`,
+						cmd,
+					];
+					if (response.error) {
+						msg.push(
+							`Status code: ${response.statusCode}`,
+							`Error: ${response.error}`,
+						);
+					} else {
+						msg.push(`Response: ${strData}`);
+					}
+					msg.push('');
+					reject(new Error(msg.join('\n')));
+				}
 			});
-			this.stdin.write(JSON.stringify({ command, parameters }) + '\n');
+			this.stdin.write(cmd);
 		});
 	}
 
--- a/src/preload.py
+++ b/src/preload.py
@@ -3,7 +3,7 @@ import json
 import os
 import re
 import sys
-
+import traceback
 
 from contextlib import contextmanager
 from functools import partial
@@ -649,7 +649,8 @@ def start_dockerd_and_wait_for_stdin(app_data, image=None):
         write_apps_json(app_data, mpoint + "/apps.json")
         with docker_context_manager(driver, mpoint):
             # Signal that Docker is ready.
-            print(json.dumps({}))
+            print(json.dumps({"statusCode": 0}))
+            sys.stdout.flush()
             # Wait for the js to finish its job.
             input()
 
@@ -955,8 +956,20 @@ methods = {
 if __name__ == "__main__":
     update_ca_certificates(**SH_OPTS)
     for line in sys.stdin:
-        data = json.loads(line)
-        method = methods[data["command"]]
-        result = method(**data.get("parameters", {}))
-        print(json.dumps({"result": result}))
+        try:
+            data = json.loads(line)
+            method = methods[data["command"]]
+            result = method(**data.get("parameters", {}))
+            response = {
+                "result": result,
+                "statusCode": 0,
+            }
+        except BaseException:
+            response = {
+                "error": traceback.format_exc(),
+                "statusCode": 1,
+            }
+        print(json.dumps(response))
         sys.stdout.flush()
+        if response["statusCode"]:
+            break


src/preload.py
```diff
@@ -649,7 +649,8 @@ def start_dockerd_and_wait_for_stdin(app_data, image=None):
         write_apps_json(app_data, mpoint + "/apps.json")
         with docker_context_manager(driver, mpoint):
             # Signal that Docker is ready.
-            print(json.dumps({}))
+            print(json.dumps({"statusCode": 0}))
+            sys.stdout.flush()
```
pdcastro, line 653: This is uncomfortable, but it seems to work OK. Unfortunately, I found that there was additional communication between the CLI process and the Python process outside the `_runCommand` "framework", namely the `print(...)` statement above on the Python side, and the following `this.stdin.write('\n');` on the CLI side:

https://github.com/balena-io-modules/balena-preload/blob/v10.4.4/lib/preload.js#L1060-L1061

src/preload.py
```diff
@@ -955,8 +956,20 @@ def get_partition(name, image=None):
...
-        print(json.dumps({"result": result}))
+        try:
+            data = json.loads(line)
+            method = methods[data["command"]]
+            result = method(**data.get("parameters", {}))
+            response = {
+                "result": result,
+                "statusCode": 0,
+            }
+        except BaseException:
```
pdcastro, line 967: Alexis suggested that `Exception` should be sufficient, but I'm still minded on `BaseException` because of this hierarchy list:
https://docs.python.org/3.6/library/exceptions.html#exception-hierarchy

Basically I wanted to ensure that absolutely any exception will be reported to the CLI process, causing it to terminate, to avoid the bad UX of infinite spinning wheels.

lib/preload.js
```diff
@@ -382,13 +382,56 @@ class Preloader extends EventEmitter {
...
+	 */
 	_runCommand(command, parameters) {
 		return new Bluebird((resolve, reject) => {
+			const cmd = JSON.stringify({ command, parameters }) + '\n';
 			this.stdout.once('error', reject);
 			this.stdout.once('data', (data) => {
-				resolve(JSON.parse(data).result);
+				let strData = data;
+				try {
+					strData = data.toString();
```
pdcastro, line 404: `data` is normally a `Buffer` object, not a string. `JSON.parse(data)` would implicitly call `data.toString()`, now I'm making it explicit so it can also be used in error messages.

lib/preload.js
```diff
@@ -382,13 +382,56 @@ class Preloader extends EventEmitter {
...
+						`An error has occurred executing internal preload command '${command}':`,
+						cmd,
+					];
+					if (response.error) {
+						msg.push(
+							`Status code: ${response.statusCode}`,
+							`Error: ${response.error}`,
+						);
+					} else {
+						msg.push(`Response: ${strData}`);
```
pdcastro, line 428: If `response.error` is set, it's a case of "graceful" error reporting from the Python side. If it is not set, something unexpected happened, so we just collect the raw `strdData`, hoping that it helps with debugging.

pdcastro:
There is still a problem, not necessarily to be fixed in this PR, but some warning messages -- that may turn out to be actual errors -- are still hidden behind infinite spinning wheels, unless `--debug` or `DEBUG=1` is used. These warnings:

```
Retrying (count=2) /sbin/losetup -f --show /dev/loop8 False
Hint: If using a Virtual Machine, consider increasing the number of processors.
If using Docker Desktop for Windows or macOS, it may require restarting.
| Reading image information

Retrying (count=3) /sbin/losetup -f --show /dev/loop8 False
Hint: If using a Virtual Machine, consider increasing the number of processors.
If using Docker Desktop for Windows or macOS, it may require restarting.
/ Reading image information
...
```
I have to restart Docker Desktop to get out of that condition, but without `--debug` I am only shown this infinite spinning:

```
| Reading image information
/ Reading image information
- Reading image information
```

Some end users will wait for hours at this spinning, thinking that the CLI is hard at work, when actually it is stuck.

lib/preload.js
```diff
@@ -382,13 +382,56 @@ class Preloader extends EventEmitter {
...
+	 * assumed to have failed and the returned promise is rejected with an
+	 * error message including the message provided in the `error` field.
+	 *
+	 * @param {string} command
+	 * @param {object} parameters
+	 */
 	_runCommand(command, parameters) {
 		return new Bluebird((resolve, reject) => {
+			const cmd = JSON.stringify({ command, parameters }) + '\n';
 			this.stdout.once('error', reject);
```
klutchell, line 400: Can you confirm my thinking that dockerode/Bluebird communication errors (not related to our python script) are caught by this `('error', reject)` function? It looks good to me and that's why we don't want to mix up our in-band `statusCode` with the out-of-band `StatusCode`.

lib/preload.js
```diff
@@ -382,13 +382,56 @@ class Preloader extends EventEmitter {
...
+				if (response.statusCode === 0) {
+					resolve(response.result);
+				} else {
+					const msg = [
+						`An error has occurred executing internal preload command '${command}':`,
+						cmd,
+					];
+					if (response.error) {
+						msg.push(
+							`Status code: ${response.statusCode}`,
```
klutchell, line 424: I was under the impression that something in the CLI was explicitly looking for `StatusCode` in a failed preload command, so this is the only place where we actually want our python graceful failure to mimic that of a dockerode failure. I might just not be clear on how we are using the reject errors.

lib/preload.js
pdcastro: My understanding is limited -- including my understanding of your question, 😁   not sure exactly what to make of "dockerode/Bluebird communication errors". I am tempted to say that the answer is 'no'. Focusing specifically this line:
```js
	this.stdout.once('error', reject);
```
Through this line, `reject` will be called on errors emitted by the `this.stdout` stream (obviously). And in fact only the very first error (that gets emitted after the `reject` handler is attached), because the line uses `.once()` instead of `.on()`. Going back to where `this.stdout` is attached:
https://github.com/balena-io-modules/balena-preload/blob/v10.4.4/lib/preload.js#L879
```js
	this._prepareErrorHandler();
	return this.container.attach({ stream: true, stdout: true, stderr: true, stdin: true });
}).then((stream) => {
	this.stdin = stream;
	this.docker.modem.demuxStream(stream, this.stdout, this.stderr);
```
So I gather that `this.stdout` is the stdout of the container where the Python script is running. However, looking at the source source of `docker-modem` (v1.0.9 as actually currently used by `balena preload`), the implementation of `demuxStream`:
https://github.com/apocas/docker-modem/blob/v1.0.9/lib/modem.js#L304
```
Modem.prototype.demuxStream = function(stream, stdout, stderr) {
	...
	stdout.write(payload);
```
... we see there is no `.pipe(stdout)` connecting a "source of error events" to `this.stdout`, only the occasion `stdout.write(payload)` where `payload` is probably a `Buffer` object.

So I think that `this.stdout.once('error', reject);` might never actually produce any errors, `reject` might never actually be called on that line.

But so what? :-)  Why does it matter?

Note that the are other rejection routes at play. The high level `prepare()` and `preload()` functions in `preload.js` return promises that may reject at various points, and also the `Preloader` class itself is a subclass of `EventEmitter`:

```js
    class Preloader extends EventEmitter {
```

At the CLI level, there is the following line:

https://github.com/balena-io/balena-cli/blob/v12.44.12/lib/commands/preload.ts#L285
```js
    preloader.on('error', reject);
```

An in PR #241, I added a missing `this.emit()`:

https://github.com/balena-io-modules/balena-preload/pull/241/files#diff-337c0a7cc97920afbd5531862b39c73752b1377bbda24c26add1e511c63a2cc6R382
```
   .catch((error) => this.emit('error', error));
```

In `preload.js`, there are now 2 occurrences of `this.emit('error', error);`, and if any is called, the CLI should catch it at the top level.

I am not stating that we have covered all possible sources of errors. When you mentioned "dockerode/Bluebird communication errors", what specific `dockerode` call did you have in mind?

lib/preload.js
pdcastro: `statusCode` is not inspected anywhere else, only in this block of code. This error message is just a string that may be printed to console when the promise rejection bubbles up. It is just a coincidence that the name `statusCode` looks similar to `StatusCode` in the following lines:

```js
		this.container.wait().then((data) => {
				if (data.StatusCode !== 0) {
```

My motivation / rationale was to try and detect, around here, whether the Python script was successful in executing a "preload command" (JSON object written to the Python script's `stdin`). For this purpose, I insist on `response.statusCode === 0` being true (note the use of `===` instead of `==`, so I'm looking for "number zero" specifically, not any falsy value like empty string, null, undefined).

src/preload.py
```diff
@@ -955,8 +956,20 @@ def get_partition(name, image=None):
...
+            response = {
+                "error": traceback.format_exc(),
+                "statusCode": 1,
+            }
+        print(json.dumps(response))
         sys.stdout.flush()
+        if response["statusCode"]:
+            break
```
pdcastro, line 975: This `break` means that the Python process should exit when the first error is encountered, rather than continue running and waiting for the next command on `stdin`. I thought this makes sense because, on the CLI / `preload.js` side, I am desperately trying to ensure that the CLI process exits on detecting any errors (to solve the poor UX of infinite spinning wheels).

lib/preload.js
klutchell: Oh I see, Bluebird is just a promise library. I thought it was part of the dockerode communication module. In that case it makes sense that `this.stdout.once('error', reject);` may never get called. I even tried returning data on stderr in the python script and it still triggered the `data` handler and not the `error` handler.

klutchell (approved):
I just tested this branch locally and in a VM and it looks good to me (fails without aufs and succeeds for overlay2).

pdcastro:
@klutchell, thank you for your reviews 👍 
FYI, I have a commit in CLI PR balena-io/balena-cli/pull/2259 to bump `balena-preload` alongside other dependencies. I am currently testing it.

PR #241, opened 2021-05-04
Fix unhandled exception on container.wait()

pdcastro:
Connects-to: #233

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -360,23 +360,26 @@ class Preloader extends EventEmitter {
 
 	_prepareErrorHandler() {
 		// Emit an error event if the python script exits with an error
-		this.container.wait().then((data) => {
-			if (data.StatusCode !== 0) {
-				const output = this.bufferedStderr.getData().toString('utf8').trim();
-				let error;
-				if (
-					output.indexOf(GRAPHDRIVER_ERROR) !== -1 &&
-					output.indexOf(DOCKERD_USES_OVERLAY) !== -1
-				) {
-					error = new this.balena.errors.BalenaError(OVERLAY_MODULE_MESSAGE);
-				} else {
-					error = new Error(output);
-					// @ts-ignore
-					error.code = data.StatusCode;
+		this.container
+			.wait()
+			.then((data) => {
+				if (data.StatusCode !== 0) {
+					const output = this.bufferedStderr.getData().toString('utf8').trim();
+					let error;
+					if (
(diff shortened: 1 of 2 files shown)


lib/preload.js
```diff
@@ -24,7 +24,6 @@ const DOCKER_IMAGE_TAG = 'balena/balena-preload';
 const DISK_IMAGE_PATH_IN_DOCKER = '/img/balena.img';
 const SPLASH_IMAGE_PATH_IN_DOCKER = '/img/balena-logo.png';
 const DOCKER_STEP_RE = /Step (\d+)\/(\d+)/;
-const EXIT_CAUSED_BY_SIGINT = 137;
```
pdcastro, line 27: I gather that code 137 is for SIGKILL, not SIGINT, and it could be caused by the container / system running out of memory. I assume we would want to report OOM SIGKILL through to the CLI process, hence removing the test `data.StatusCode !== EXIT_CAUSED_BY_SIGINT`.

lib/preload.js
```diff
@@ -361,23 +360,26 @@ class Preloader extends EventEmitter {
...
+						// @ts-ignore
+						error.code = data.StatusCode;
+					}
+					this.emit('error', error);
 				}
-				this.emit('error', error);
-			}
-		});
+			})
+			.catch((error) => this.emit('error', error));
```
pdcastro, line 382: This `.catch` was missing. The only changes in this whole block are adding the last `.catch` line and removing `&& data.StatusCode !== EXIT_CAUSED_BY_SIGINT`. The rest was changed by a moody linter. :-)

package.json
```diff
@@ -60,6 +60,7 @@
   "scripts": {
     "lint": "balena-lint -e js --typescript lib",
     "lint-fix": "balena-lint -e js --typescript --fix lib",
+    "lint-python": "flake8 src/preload.py",
```
pdcastro, line 63: This `npm` script was added as a "hint" for someone to execute it manually if they know and remember about it. Alexis mentioned it to me in this older PR: https://github.com/balena-io-modules/balena-preload/pull/220#pullrequestreview-446302471

Automated execution is a "To Do". The `flake8` command is a python utility, not `npm` package, so cannot (?) be installed with something like "npm install", and adding some sort of "pip install" would mean having to deal with setting up a python environment or Dockerfile.

pdcastro:
@balena-ci I self-certify!

PR #242, opened 2021-05-07
Append to existing apps.json preload state instead of overwriting

alexgg:
This is to remain in draft state until the v3 target state API changes are merged

-----

If the image we are preloading to contains an existing `apps.json` append
the preloaded app to it instead of overwriting it.

Fixes balena-io/balena-cli#2262

Relates to balena-io/balena-cli#2263

Change-type: patch
Signed-off-by: Alex Gonzalez <alexg@balena.io>

------

Tested on a `raspberrypi4-64-2.75.0+rev1` image that does not contain a preloaded `apps.json`, and on a genericx86-64-ext development build with an existing `apps.json` with preloaded state.

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -466,7 +466,7 @@ class Preloader extends EventEmitter {
 				return this.balena.request.send({
 					headers: { 'User-Agent': SUPERVISOR_USER_AGENT },
 					baseUrl: this.balena.pine.API_URL,
-					url: `device/v${this._supervisorLT7() ? 1 : 2}/${uuid}/state`,
+					url: `device/v${this._supervisorLT7() ? 1 : this._supervisorLT13() ? 2 : 3}/${uuid}/state`,
 				});
 			})
 			.get('body')
@@ -735,6 +735,15 @@ class Preloader extends EventEmitter {
 			return compareVersions(this.supervisorVersion, '7.0.0') === -1;
 		} catch (e) {
 			// Suppose the supervisor version is >= 7.0.0 when it is not valid semver.
+			return false;
+		}
+	}
+
+	_supervisorLT13() {
+		try {
+			return compareVersions(this.supervisorVersion, '13.0.0') === -1;
+		} catch (e) {
+			// Suppose the supervisor version is >= 13.0.0 when it is not valid semver.
 			return false;
 		}
 	}


src/preload.py
```diff
@@ -608,13 +608,14 @@ def write_resin_device_pinning(app_data, output):
             ),
         )
 
-
```
pdcastro, line 611: Re deleted blank lines: if you `npm run lint-python`, which in turn runs `flake8`, I think you will find that the linter will put the blank lines back. :-) I think it is from PEP-8:
> https://www.python.org/dev/peps/pep-0008/#blank-lines
> Blank Lines
> Surround top-level function and class definitions with two blank lines.

src/preload.py
```diff
@@ -608,13 +608,14 @@ def write_resin_device_pinning(app_data, output):
...
 def write_apps_json(data, output):
-    """Writes data dict to output as json"""
+    """Updates or creates preloaded state"""
+    if os.path.isfile(output):
+        with open(output, "r") as f:
+            data['apps'].update(json.load(f)['apps'])
```
pdcastro, line 615: What if the image file had already been preloaded, e.g. if the `balena preload` is executed a second time for the same image file? The earlier execution would take precedence, right? It should be the other way around, the latter execution taking precedence.

Actually, I think it is preferable to forbid executing `balena preload` on the same image file multiple times, because I think users assume that subsequent executions will clean up / undo previous executions, when in fact it doesn't and produces a bloated, unnecessarily large image file. If we could reliably detect that `balena preload` was executed before for this same file (preferably by inspecting the image file itself), I think this would be best. We could print an error message along the lines of:
> Error: this image file has been preloaded before.
> Please use a fresh, not-previously-preloaded image file.

src/preload.py
pdcastro: We were typing at the same time, see my other comment :-)
Is there a valid use case for running `balena preload` multiple times for the same image file?

src/preload.py
```diff
@@ -610,7 +610,17 @@ def write_resin_device_pinning(app_data, output):
 
 
 def write_apps_json(data, output):
-    """Writes data dict to output as json"""
+    """Updates or creates preloaded state"""
+    if os.path.isfile(output):
+        with open(output, "r") as f:
+            current_state = json.load(f)
+        for uuid in data.keys():
+            if len(uuid) != len(current_state.keys()[0]):
```
pdcastro, line 618: I understand that `current_state` is the existing `apps.json` in the image file, and `data` is data loaded from the API. So in the expression `for uuid in data.keys()`, currently/today, `uuid` is actually a release ID, right? The release ID being a number like 12345678, rather than a UUID.

To confirm my understanding:
* Future PRs to `balena-preload` (or `balena-cli` or both) will be required for `uuid` to actually be a UUID (as it is currently a release ID).
* Before this PR, the CLI could not be used with new hostapp-aware images because `apps.json` would be overwritten (not because of ID vs UUID, which is a supervisor matter).
* After this PR, the CLI still can be used with new hostapp-aware images because `apps.json` will be merged, even if release IDs (rather than UUIDs) are still used.

pdcastro: Old images that are not hostapp aware don't have any existing `apps.json` file, or do they?
In which case, `current_state` would not be defined and `current_state.keys()` would raise an exception?

src/preload.py
```diff
@@ -610,7 +610,17 @@ def write_resin_device_pinning(app_data, output):
...
+                raise Exception("Current and target preloaded states not "
+                                "compatible")
+            if uuid not in current_state['apps']:
```
pdcastro, line 621: But this would mean that running preload consecutive times (on the same, already-preloaded image) would not overwrite previous entries of `apps.json`, which is not good UX (unexpected behaviour from users' point of view).

If you simply delete this `if` clause and always run `current_state['apps'][uuid] = data[uuid]`, this might work better? I understand we don't want to wipe a pre-existing `apps.json`, but we still want to overwrite entries where the `uuid` matches, don't we?

src/preload.py
alexgg: @pdcastro I was actually thinking about a multi-app future where users would  run preload several times to load different apps, that's why I implemented it that way.

src/preload.py
alexgg: @pdcastro in `for uuid in data.keys()`, data is of the form:
```
{
    "apps": {
        "119711": {
            "commit": "d99be5beb5cffd74af068cb4f4f55389",
            "name": "rpi4_idle",
            "networks": {},
            "releaseId": 133467,
```
So the `uuid` is currently the appID (119711), not the releaseID (133467).

From your points:

* Future PRs to balena-preload (or balena-cli or both) will be required for uuid to actually be a UUID (as it is currently an app ID). This change is also happening in the UI for a v3 target state.
* Before this PR, the CLI could not be used with new hostapp-aware images because apps.json would be overwritten, and also because the SV required in those images require apps.json to use appUUID and not appID.
After this PR, the CLI still cannot be used with new hostapp-aware images because the preloaded app does not use appUUIDs.

I will leave this as a draft until we have the whole picture.

src/preload.py
alexgg: There is a check for the existence of `apps.json` so if it is not there the behaviour is just as it was before.
```
if os.path.isfile(output):
```
I really don't like the idea of the CLI modifying the current state that exactly matches what the OS was built with. It should just add apps but not touch existing apps, specially hostos apps as it might break the system.

pdcastro: > Why would you want to overwrite entries?

Here's the scenario:

1. Run `balena preload` and select a certain release "R1" of user application "A1".
1.1 The image file now contains an `apps.json` file that lists the selected release "R1" for app "A1".
2. _"Oh wait! I've preloaded the wrong release, duh. I meant to preload release "R2", not release "R1". Let me run it again."_
3. Run `balena preload` again for the same image file and same app, but this time select release "R2".
3.1 Expected result: the image file (apps.json) now lists release "R2" for app "A1".
3.2 Actual, unexpected result: the image file (apps.json) still lists release "R1" for app "A1", because the code above refuses to overwrite it.

All the above happening before the image is flashed, which means we were not on the same page when you wrote _"why wear the media with a write?"_ I still haven't made sense of this question. :-)

src/preload.py
```diff
@@ -609,8 +609,42 @@ def write_resin_device_pinning(app_data, output):
...
+            # Until supervisor is part of the hostOS app, there is no way to
+            # distinguish it from a single service user app from the state data
+            # The only way is checking the actual image contents
+            if len(app['services']) == 1:
+                for service in app['services']:
+                    image = app['services'][service]['image']
+                version = app['releaseVersion']
+                if isSupervisorApp(image, version) is True:
+                    continue
+            del current_state['apps'][uuid]
```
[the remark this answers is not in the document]
alexgg: > It looks like we can distinguish between host OS apps, supervisor apps and "user apps" 
We can cleanly distinguish user apps from hostOS app with the `isHost` attribute, which works once the supervisor becomes a hostOS app. Right now distinguishing the supervisor app from user apps is ugly - I need to read the image info and do a regex with the image name. Given that we know how this will be cleanly worked out in future I guess we can live with the hack.

>  when a user runs balena preload a second time for the same image file, preloading a different release the second time around, the contents of both releases are saved to the image file

That's correct. I have to say that after looking at this code it looks over-engineered and sub-optimal.

 > I would suggest that instead of deleting existing entries from apps.json (del current_state['apps'][uuid]), we threw an error 

That's a UX decision, I don't know how preloading is currently used. Are users version controlling a raw image and then updating the preloaded app as time goes by? Will asking them to always re-download the raw image be a problem? I am guessing not.

> Would balena preload be used to preload or replace host apps or supervisor apps in addition to "user apps"

Nothing has been discussed about this, and I expect not to be the case until we decide to expose hostOS apps to users which is not currently the case.

When that happens we can probably come up with a much cleaner implementation of the preloading engine that receives current state as input, prepares a data partition image to match, and writes this image over the data partition in the raw file. This could be the same transformation that happens at OS build time, and it could be reused in the cloud if required.

src/preload.py
```diff
@@ -609,8 +609,42 @@ def write_resin_device_pinning(app_data, output):
...
+        current_state['apps'][uuid] = data[uuid]
```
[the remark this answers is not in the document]
alexgg: > data[uuid] or data['apps'][uuid]

You were too quick, I had pushed at my end-of-day and did not test until today. I have pushed the modified code now.

> By the way, other than "apps", what are the other top-level sections? 

There is the `pinDevice` and the `config` sections that I know about. I am trying to understand the `config` section better, it's not something that the OS adds to the preloaded `apps.json` and I want to understand what is its purpuse. For the time being I am just appending both `pinDevice` and `config` to the final state.

src/preload.py
alexgg: I am distinguishing supervisor from user apps by reading the image names and comparing with the current state. It's ugly, but in future when the supervisor becomes part of the hostOS it can be distinguished with the `isHost` attribute with the rest of the hostOS.

I will raise an exception if trying to re-preload as it seems a better solution.

src/preload.py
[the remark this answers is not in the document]
alexgg: I pushed the modification to exit when trying to re-preload.

src/preload.py
```diff
@@ -609,8 +609,64 @@ def write_resin_device_pinning(app_data, output):
...
-    """Writes data dict to output as json"""
+    """Updates or creates preloaded state"""
+    if os.path.isfile(output):
+        with open(output, "r") as f:
+            current_state = json.load(f)
+
+        # Check state version consistency
+        for uuid in data.keys():
+            if len(uuid) != 32 and len(uuid) != 62:
+                raise Exception("Refusing to preload v2 target state app")
```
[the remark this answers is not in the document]
alexgg: I was wondering about this while looking into how the app state is passed to the preloader code. The `preload.js` will check the supervisor version in the image and fetch a versioned target state for the app depending on the supervisor version, so there should never be a mismatch here.
This check should disappear as it will never be hit, or if we want to be paranoid check that the id length of both states are the same, and point the user to contact support if they are not.

src/preload.py
```diff
@@ -609,8 +609,62 @@ def write_resin_device_pinning(app_data, output):
...
+        images, supervisor_version, balena_os_version = (
+            get_images_and_supervisor_version()
+        )
+
+        for uuid in list(current_state['apps'].keys()):
+            # Check state version consistency
+            if len(uuid) != len(list(data['apps'].keys())[0]):
+                log.error("The preloaded app state and the current state in"
+                          "the image have a version mismatch. This is likely"
+                          "a bug, please contact Balena's support")
```
pdcastro, line 644: Could this error happen if:
* The user preloads an image file using an old version of the CLI, which overwrites `apps.json` in the image. The user saves a copy of the preloaded image file.
* "6 months later", the user upgrades the CLI to the latest version / this PR's version.
* The user attempts to preload the same image file again (not starting with a fresh image file) with the new CLI version.

Instead of "please contact Balena's support", how about:
```suggestion
                log.error("The preloaded app state and the current state in"
                          "the image have a version mismatch. Please use a fresh,"
                          "not-previously-preloaded image file.")
```
If the error still happens with a fresh image, the user would likely reach out to balena support or forums anyway. The CLI will also print this standard message when things fail:

```
$ balena preload ...
...
Additional information may be available with the `--debug` flag.

For further help or support, visit:
https://www.balena.io/docs/reference/balena-cli/#support-faq-and-troubleshooting
```

alexgg: Makes sense, I'll change it.

src/preload.py
```diff
@@ -609,8 +609,62 @@ def write_resin_device_pinning(app_data, output):
...
+        # Supervisor is single service
+        return False
+
+    # Fetch image name from state app
+    for service in app['services']:
+        image_name = app['services'][service]['image']
+
+    for image in images:
+        if re.match(image, image_name):
+            return True
```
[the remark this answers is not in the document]
alexgg: Both will do but your suggestion is probable safer. The image is an `image_location` and the image_name is an `image_location@contents_hash`, like:

```
image: registry2.balena-staging.com/v2/cb16e03ffa246c5b5a00b00f3385f4e8
image_name: registry2.balena-staging.com/v2/cb16e03ffa246c5b5a00b00f3385f4e8@sha256:b48d6909e2ab78f23284d8840faa45dca8a3b46a747e2dd958e20022ef21a562
```

alexgg:
Validation:
* On a devenv with a draft v3 state endpoint:
  [ ] Preloading on an image with preloaded `apps.json` with a v3 state supervisor
    * v3 style app is appended to apps.json
    * However preloading fails with a certification error 
    ```
    | Cleaning up temporary files
    (HTTP code 500) server error - Get https://registry2.example.com/v2/: x509: certificate signed by unknown authority
    ```
  [ ] Preloading on an image with preloaded `apps.json` with a v2 state supervisor (should not happen)
  [ ] Preloading on an image without `apps.json` with a v2 state supervisor
  [ ] Preloading on an image without `apps.json` with a v3 state supervisor (should not happen)

pdcastro (commented):
Small things while it is still a draft -- I'd have approved the PR if it wasn't a draft. :-)

PR #244, opened 2021-05-17
Avoid hardcoded registry2 URL

klutchell:
Instead we can extract the registry URL from the image location
as we do for the manifest and layer endpoints.

Resolves: https://github.com/balena-io-modules/balena-preload/issues/236
Change-type: patch

### DIFF
--- a/.resinci.yml
+++ b/.resinci.yml
@@ -1,3 +1,11 @@
 ---
 docker:
   publish: false
+npm:
+  platforms:
+    - name: linux
+      os: ubuntu
+      architecture: x86_64
+      node_versions:
+        - "10"
+        - "12"
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -514,7 +514,8 @@ class Preloader extends EventEmitter {
 		// List images from the release.
 		const images = this._getRelease().contains__image.map((ci) => {
 			return _.merge({}, ci.image[0], {
-				is_stored_at__image_location: ci.image[0].is_stored_at__image_location.toLowerCase(),
+				is_stored_at__image_location:
+					ci.image[0].is_stored_at__image_location.toLowerCase(),
 			});
 		});
 		// App from the state endpoint (v1 or v2 depending on the supervisor version).
@@ -551,6 +552,7 @@ class Preloader extends EventEmitter {
 	}
 
 	async registry(
+		registryUrl,
 		endpoint,
 		registryToken,
 		headers,
@@ -561,7 +563,7 @@ class Preloader extends EventEmitter {
 		headers = { ...headers };
 		headers['Authorization'] = `Bearer ${registryToken}`;
 		return await request({
-			url: `https://${this.balenaSettings.registry2Url}${endpoint}`,
+			url: `https://${registryUrl}${endpoint}`,
 			headers: headers,
 			json: decodeJson,
 			simple: false,
@@ -571,7 +573,7 @@ class Preloader extends EventEmitter {
 		});
 	}
 
-	async _getLayerSize(token, layerUrl) {
+	async _getLayerSize(token, registryUrl, layerUrl) {
 		// This gets an approximation of the layer size because:
 		// * it is the size of the tar file, not the size of the contents of the tar file (the tar file is slightly larger);
 		// * the gzip footer only gives the size % 32 so it will be incorrect for layers larger than 4GiB
@@ -580,6 +582,7 @@ class Preloader extends EventEmitter {
 		// We don't want to send the registry token to s3 so we ask it to not follow
 		// redirects and issue the second request manually.
 		let response = await this.registry(
+			registryUrl,
 			layerUrl,
 			token,
 			headers,
@@ -603,6 +606,12 @@ class Preloader extends EventEmitter {
 			);
 		}
 		return response.body.readUIntLE(0, 4);
+	}
+
+	_registryUrl(imageLocation) {
+		// imageLocation: registry2.balena-cloud.com/v2/ad7cd3616b4e72ed51a5ad349e03715e@sha256:4c042f195b59b7d4c492e210ab29ab61694f490a69c65720a5a0121c6277ecdd
+		const slashIndex = imageLocation.search('/');
+		return `${imageLocation.substring(0, slashIndex)}`;
 	}
 
 	_imageManifestUrl(imageLocation) {
@@ -633,6 +642,7 @@ class Preloader extends EventEmitter {
 			imagesLocations,
 			async (imageLocation) => {
 				const { body: manifest } = await this.registry(
+					this._registryUrl(imageLocation),
 					this._imageManifestUrl(imageLocation),
 					registryToken,
 					{},
@@ -662,6 +672,7 @@ class Preloader extends EventEmitter {
 			async ({ imageLocation, layer }) => {
 				const size = await this._getLayerSize(
 					registryToken,
+					this._registryUrl(imageLocation),
 					this._layerUrl(imageLocation, layer.digest),
 				);
 				layersSizes.set(layer.digest, { size, compressedSize: layer.size });
@@ -734,7 +745,7 @@ class Preloader extends EventEmitter {
 				baseUrl: this.balena.pine.API_URL,
 				url: '/auth/v1/token',
 				qs: {
-					service: this.balenaSettings.registry2Url,
+					service: this._registryUrl(images[0]),
 					scope: images.map(
 						(imageRepository) =>
 							`repository:${imageRepository.substr(
--- a/package.json
+++ b/package.json
@@ -58,8 +58,7 @@
     "url": "https://github.com/balena-io/balena-preload/issues"
   },
   "scripts": {
-    "lint": "balena-lint -e js --typescript lib",
-    "lint-fix": "balena-lint -e js --typescript --fix lib",
+    "lint": "balena-lint -e js --typescript --fix lib",
     "lint-python": "flake8 src/preload.py",
     "test": "npm run lint",
     "prepare": "tsc"
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@types/node": "^10.17.28",
     "@types/tmp": "0.0.34",
     "@types/unzipper": "^0.8.4",
+    "catch-uncommitted": "^2.0.0",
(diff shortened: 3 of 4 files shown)


pdcastro (dismissed):

klutchell:
I'll format with `prettier` and push again but the current editorconfig does not match the 4-space tabs currently used in this file.

https://github.com/balena-io-modules/balena-preload/blob/29411dfe11dc26f23695860a5d4c771f5b9fce4b/.editorconfig#L6-L12

pdcastro:
OK 👍  (If you need to prevent the PR from auto merging, you can add the `versionbot/pr-draft` label.)

lib/preload.js
```diff
@@ -561,7 +563,7 @@ class Preloader extends EventEmitter {
 		headers = { ...headers };
 		headers['Authorization'] = `Bearer ${registryToken}`;
 		return await request({
-			url: `https://${this.balenaSettings.registry2Url}${endpoint}`,
+			url: `https://${registryUrl}${endpoint}`,
```
pdcastro, line 566: In addition to this one, there is another occurrence of `registry2Url` that I think needs amending:
https://github.com/balena-io-modules/balena-preload/blob/v10.4.6/lib/preload.js#L737

pdcastro (changes requested):

pdcastro:
Nice 👍   @klutchell, did you get to test this PR with openBalena, to confirm that the registry2 error happens before this PR, and does not happen after this PR? If not, could you test it? To make this kind of testing easier, I've put together these instructions and scripts: https://github.com/pdcastro/ob_multipass
(I haven't shared it more widely because it's very opinionated about the user of multipass and everything else, and also because it is not sufficient to produce an installation of openBalena to which devices can be added. But I found it handy to test CLI issues, and in this case, `preload`.)

klutchell:
I've only tested with balena-cloud.com so far, just to ensure the URL parsing was working. I'll give it a shot with your instructions above and then reopen this PR.

.resinci.yml
```diff
@@ -1,3 +1,19 @@
...
+        - "10"
+        - "12"
+        - "14"
+    - name: linux
+      os: alpine
+      architecture: x86_64
+      node_versions:
+        - "10"
+        - "12"
+        - "14"
```
pdcastro, line 19: > CI: enable CI for the same platforms as balena-cli

That's fair, but... Considering that the `balena-preload` tests are currently limited to `npm run lint`, we could actually have a `.resinci.yml` more like [the one from balena-multibuild](https://github.com/balena-io-modules/balena-multibuild/blob/v4.11.0/.resinci.yml):
```
npm:
  platforms: []
```
Alternatively, we could add `catch-uncommitted` to the `test` npm script -- see other comment -- so that at least there is a chance that the build could fail. Would probably not need to test it in alpine, ubuntu and 3 different versions of node though. 😁

package.json
```diff
@@ -58,8 +58,7 @@
     "url": "https://github.com/balena-io/balena-preload/issues"
   },
   "scripts": {
-    "lint": "balena-lint -e js --typescript lib",
-    "lint-fix": "balena-lint -e js --typescript --fix lib",
+    "lint": "balena-lint -e js --typescript --fix lib",
     "lint-python": "flake8 src/preload.py",
     "test": "npm run lint",
```
pdcastro, line 63: Given that we are so serious about linting that it's the only thing we care about in our automated testing, :-) we might as well make sure that the build fails if the linter produces uncommitted changes, as follows:
```suggestion
    "test": "npm run lint && catch-uncommitted --skip-node-versionbot-changes",
```
Plus adding `"catch-uncommitted": "^2.0.0",` to the `devDependencies` section of `package.json`.
Ref: https://github.com/balena-io-modules/catch-uncommitted/
I expect this would work, but I am not sure...

.resinci.yml
klutchell: I tried to enable only alpine node 10 but I guess ubuntu 10 & 12 are required by the project so I can stick to those for now.

klutchell:
I can't seem to reproduce the original issue with OB 3.2.1 and CLI 12.37.0.

It seems to always use `registry.open.balena` and succeed.

```bash
ubuntu@localmachine:~$ balena-cli/balena settings 
balenaUrl:            open.balena
apiUrl:               https://api.open.balena
vpnUrl:               vpn.open.balena
registryUrl:          registry.open.balena
registry2Url:         registry2.open.balena
imageMakerUrl:        https://img.open.balena
deltaUrl:             https://delta.open.balena
dashboardUrl:         https://dashboard.open.balena
proxyUrl:             devices.open.balena
dataDirectory:        /home/ubuntu/.balena
projectsDirectory:    /home/ubuntu/BalenaProjects
cacheDirectory:       /home/ubuntu/.balena/cache
binDirectory:         /home/ubuntu/.balena/bin
imageCacheTime:       604800000
tokenRefreshInterval: 3600000
apiKeyVariable:       BALENA_API_KEY

ubuntu@localmachine:~$ balena-cli/balena version
12.37.0
```

Still investigating how this is possible...

pdcastro:
> It seems to always use `registry.open.balena` and succeed.

Just some informed guesswork:

* Could it be that it is using `registry2.open.balena`, but somehow that resolves to something that works? (Maybe an entry exists in the relevant `/etc/hosts`?)
* Some of the queries related to `registry2` are for pull operations, and perhaps if cached images already exist locally to the docker engine, the pull is somewhat skipped. In this case, try `docker rmi` for some relevant images, if you can identify them.
* Maybe we need different versions of something to reproduce it? This is one of the original reports:  
https://forums.balena.io/t/balena-preload-points-to-wrong-registry/260600

klutchell:
> but somehow that resolves to something that works?

I already confirmed that registry2.open.balena does not resolve to anything via curl and ping.

> perhaps if cached images already exist locally to the docker engine, the pull is somewhat skipped

Good thought, but I cleared all locally cached images and layers between tests.

I'm beginning to think that we are querying the registry URL from the API, based on the following snippets but I'm not clear on what the `qs` part of that request does yet.

https://github.com/balena-io-modules/balena-preload/blob/29411dfe11dc26f23695860a5d4c771f5b9fce4b/lib/preload.js#L737

https://github.com/balena-io/open-balena-registry/blob/ff39362394b2a6c9788c68e22c762fba6ba8611b/config/confd/templates/docker-registry.yml.tmpl#L17-L22

https://github.com/balena-io/open-balena/blob/071b5850a9a4dc6e987c6a068e35290d0ade5d03/compose/services.yml#L70-L72

pdcastro:
> It seems to always use `registry.open.balena` and succeed.

So if you add the following:
```js
    console.error(`preload settings:\n${JSON.stringify(balenaSettings, null, 4)}`);
```
to this line: https://github.com/balena-io-modules/balena-preload/blob/v10.4.6/lib/preload.js#L853

... you're saying that you get this:
```
    "registryUrl": "registry.open.balena",
    "registry2Url": "registry.open.balena",
```
instead of this?
```
    "registryUrl": "registry.open.balena",
    "registry2Url": "registry2.open.balena",
```

If the changes in this PR still work with openBalena and balenaCloud, we can merge the PR, produce a new CLI release and ask the forum users to test it (explaining we're unsure of the solution because we couldn't reproduce the error) and let us know whether it works for them.

klutchell:
The settings are what we expect, different values for registry and registry2, but I still can't reproduce the failure. I'm going to add more debugging around my changes before we go further.

```
preload settings:
{
    "balenaUrl": "open.balena",
    "apiUrl": "https://api.open.balena",
    "vpnUrl": "vpn.open.balena",
    "registryUrl": "registry.open.balena",
    "registry2Url": "registry2.open.balena",
    "imageMakerUrl": "https://img.open.balena",
    "deltaUrl": "https://delta.open.balena",
    "dashboardUrl": "https://dashboard.open.balena",
    "proxyUrl": "devices.open.balena",
    "tunnelUrl": "tunnel.open.balena",
    "dataDirectory": "/home/ubuntu/.balena",
    "projectsDirectory": "/home/ubuntu/BalenaProjects",
    "cacheDirectory": "/home/ubuntu/.balena/cache",
    "binDirectory": "/home/ubuntu/.balena/bin",
    "imageCacheTime": 604800000,
    "tokenRefreshInterval": 3600000,
    "apiKeyVariable": "BALENA_API_KEY"
}
```

lib/preload.js
```diff
@@ -734,7 +745,7 @@ class Preloader extends EventEmitter {
 				baseUrl: this.balena.pine.API_URL,
 				url: '/auth/v1/token',
 				qs: {
-					service: this.balenaSettings.registry2Url,
+					service: this._registryUrl(images[0]),
```
klutchell, line 748: This will return a valid token even if registry2 isn't a reachable URL. Still good to use a valid endpoint though!

lib/preload.js
```diff
@@ -571,7 +573,7 @@ class Preloader extends EventEmitter {
 		});
 	}
 
-	async _getLayerSize(token, layerUrl) {
+	async _getLayerSize(token, registryUrl, layerUrl) {
```
klutchell, line 576: Only called via `_getApplicationSize` when multiple images are pulled so size calculations  are required.

lib/preload.js
```diff
@@ -662,6 +672,7 @@ class Preloader extends EventEmitter {
 			async ({ imageLocation, layer }) => {
 				const size = await this._getLayerSize(
 					registryToken,
+					this._registryUrl(imageLocation),
```
klutchell, line 675: Only called via `_getApplicationSize` when multiple images are pulled so size calculations  are required.

lib/preload.js
```diff
@@ -633,6 +642,7 @@ class Preloader extends EventEmitter {
 			imagesLocations,
 			async (imageLocation) => {
 				const { body: manifest } = await this.registry(
+					this._registryUrl(imageLocation),
```
klutchell, line 645: Only called via `_getApplicationSize` when multiple images are pulled so size calculations  are required.

klutchell:
I found that `_getApplicationSize` and thus all of my changes (except `_getRegistryToken`) are only called when multiple images are part of the release being preloaded. I couldn't reproduce because I was testing with a single image release.

I switched to a multi-container example and reproduced the issue, and confirmed the fix in this PR.

pdcastro (approved):

PR #246, opened 2021-06-04
Fix supervisor repository regex

alexgg:
Update the supervisor repository regex so the supervisor version can be
correctly extracted from the images.

Change-type: patch
Signed-off-by: Alex Gonzalez <alexg@balena.io>

### DIFF
--- a/src/preload.py
+++ b/src/preload.py
@@ -51,7 +51,7 @@ CONFIG_PARTITIONS = [
     "flash-boot",  # flasher images
 ]
 
-SUPERVISOR_REPOSITORY_RE = "^resin(playground)?/[a-z0-9]+-supervisor$"
+SUPERVISOR_REPOSITORY_RE = "^((balena|resin|balenaplayground)/)?(armel|rpi|armv7hf|aarch64|i386|amd64|i386-nlp)-supervisor$"
 
 # 'sh' module '_truncate_exc' option:
 # http://amoffat.github.io/sh/sections/special_arguments.html#truncate-exc


src/preload.py
```diff
@@ -51,7 +51,7 @@
     "flash-boot",  # flasher images
 ]
 
-SUPERVISOR_REPOSITORY_RE = "^resin(playground)?/[a-z0-9]+-supervisor$"
+SUPERVISOR_REPOSITORY_RE = "^balena(playground)?/[a-z0-9]+-supervisor$"
```
pdcastro, line 54: Do we need to match ***either*** `resin` or `balena` instead of just one of them?
The regular expression is matched against this output:
```
            output = docker(
                "images",
                "--all",
                "--format",
                "{{.Repository}} {{.Tag}}"
            )
```
https://github.com/balena-io-modules/balena-preload/blob/v10.4.7/src/preload.py#L754

If the `docker images` command above was executed on an image of an old-enough version of balenaOS, would it contain `resin` instead of `balena`?

alexgg: You're right, versions before v2.26 will still use resin. Also, as we have moved to using the internal registry we will lose the "balena" or "resin" part in the near future, so the regex should be:
```
^(balena|resin)?(playground)?/?[a-z0-9]+-supervisor$
```

src/preload.py
```diff
@@ -51,7 +51,7 @@
     "flash-boot",  # flasher images
 ]
 
-SUPERVISOR_REPOSITORY_RE = "^resin(playground)?/[a-z0-9]+-supervisor$"
+SUPERVISOR_REPOSITORY_RE = "^(balena|resin)?(playground)?/?[a-z0-9]+-supervisor$"
```
pdcastro, line 54: This will match an invalid repo named `playground`:
```
$ python
>>> from re import match
>>> SUPERVISOR_REPOSITORY_RE = "^(balena|resin)?(playground)?/?[a-z0-9]+-supervisor$"

# unexpected - matches invalid repo 'playground'
>>> not not match(SUPERVISOR_REPOSITORY_RE, 'playground/amd64-supervisor')
True
```
A small improvement would be:
```suggestion
SUPERVISOR_REPOSITORY_RE = "^((balena|resin)(playground)?/)?[a-z0-9]+-supervisor$"
```

```
$ python
>>> from re import match
>>> SUPERVISOR_REPOSITORY_RE = "^((balena|resin)(playground)?/)?[a-z0-9]+-supervisor$"

>>> not not match(SUPERVISOR_REPOSITORY_RE, 'playground/amd64-supervisor')
False

>>> not not match(SUPERVISOR_REPOSITORY_RE, 'balena/amd64-supervisor')
True

>>> not not match(SUPERVISOR_REPOSITORY_RE, 'balenaplayground/amd64-supervisor')
True
```

With or without this improvement, it would still match some strange / invalid names as follows, but further improvement might not be worth (?) the complexity like keeping track of all valid architecture strings:
```
# missing slash
>>> not not match(SUPERVISOR_REPOSITORY_RE, 'balenaamd64-supervisor')
True

# invalid architecture
>>> not not match(SUPERVISOR_REPOSITORY_RE, 'playground-supervisor')
True
```

pdcastro (commented):
Review requested, but no changes since my last review?

> alexgg requested a review from pdcastro 2 hours ago (7 Jun 2021, 17:08 UTC)
>> pdcastro reviewed 4 hours ago (7 Jun 2021, 14:58 UTC)

src/preload.py
alexgg: @pdcastro there are not so many architectures so we can do:
```
SUPERVISOR_REPOSITORY_RE = "^(?:(?:balena|resin|balenaplayground)??/)??(?:armel|rpi|armv7hf|aarch64|i386|amd64|i386-nlp)-supervisor$"
```
Note that there is no `resinplayground` in dockerhub so I omitted it.

src/preload.py
```diff
@@ -51,7 +51,7 @@
     "flash-boot",  # flasher images
 ]
 
-SUPERVISOR_REPOSITORY_RE = "^resin(playground)?/[a-z0-9]+-supervisor$"
+SUPERVISOR_REPOSITORY_RE = "^(?:(?:balena|resin|balenaplayground)??/)??(?:armel|rpi|armv7hf|aarch64|i386|amd64|i386-nlp)-supervisor$"
```
pdcastro, line 54: I didn't get the inner `??`. It makes the expression match an empty repo such as `'/amd64-supervisor'`:
```
$ python3
>>> from re import match
>>> SUPERVISOR_REPOSITORY_RE = "^(?:(?:balena|resin|balenaplayground)??/)??(?:armel|rpi|armv7hf|aarch64|i386|amd64|i386-nlp)-supervisor$"

>>> not not match(SUPERVISOR_REPOSITORY_RE, '/amd64-supervisor')
True
```

```suggestion
SUPERVISOR_REPOSITORY_RE = "^((balena|resin|balenaplayground)/)?(armel|rpi|armv7hf|aarch64|i386|amd64|i386-nlp)-supervisor$"
```
 
Ideally we would not need to hardcode the list of possible balena architecture names here, because an architecture name could be added or renamed or aliased in the future. The SDK's `models.deviceType.getAllSupported` method could be used to get a list of "currently supported" architectures:
https://www.balena.io/docs/reference/sdk/node-sdk/#balena.models.deviceType.getAllSupported

However, "currently supported" architectures (as reported online by balena's backend) may not be exactly what we need, because a user may be trying to preload an old balenaOS image version for an architecture that we no longer support, and even downgrading the CLI would not help if the list of architectures came from balena's backend. The CLI's `balena devices supported --verbose --discontinued` command additionally uses the SDK's `models.config.getDeviceTypes` method to list discontinued device types, but the SDK has deprecated that method.
* https://www.balena.io/docs/reference/sdk/node-sdk/#balena.models.config.getDeviceTypes
* https://github.com/balena-io/balena-cli/blob/v12.44.20/lib/commands/devices/supported.ts#L89-L95

So, I don't know, maybe we go with a hardcoded list. If an architecture name is added or renamed or aliased in the future, some users would find that `balena preload` does not work as expected for their device type, they'd give up using balena or they'd contact support, GitHub issues would be created, `balena-preload` PR, `balena-cli` PR... (I understand that we don't have tests that exercise `balena preload` for all possible architectures, and even then the tests themselves should not hardcode the list of possible architectures.)

Not sure what the least bad option is. Maybe just revert to `[a-z0-9]+-supervisor` and match image names like `'not-supervisor'`? :-)

alexgg: hi @pdcastro, for me the worst option is erroneously matching something that is not the actual supervisor, as actions will be taken in that case. If we don't match for example a newer architecture we won't return a supervisor_version and won't take actions based on it.

The hardcoded list is based on https://github.com/balena-os/meta-balena/blob/51d2f658e973a999f9f346913157b7b4a6979900/meta-balena-common/recipes-containers/balena-supervisor/balena-supervisor.inc#L10.  So the OS is also hardcoding this, as it is seldom updated. 

I don't think the architectures returned by the API and the one used in supervisor names would actually match, so you would still need to keep a hardcoded table to match one with the other.

I'll update the PR with your improvement, I think that is actually the best we can do.

alexgg:
Quick validation script:

```
#!/usr/bin/python3

import re
import sys

#SUPERVISOR_REPOSITORY_RE = "^((balena|resin|balenaplayground)/?)(armel|rpi|armv7hf|aarch64|i386|amd64|i386-nlp)-supervisor$"
SUPERVISOR_REPOSITORY_RE = "^((balena|resin|balenaplayground)/)?(armel|rpi|armv7hf|aarch64|i386|amd64|i386-nlp)-supervisor$"

matches = [
    ("resin/amd64-supervisor", 1),
    ("balena/amd64-supervisor", 1),
    ("balenaplayground/amd64-supervisor", 1),
    ("amd64-supervisor", 1),
    ("balenaamd64-supervisor", 0),
    ("balena//amd64-supervisor", 0),
    ("balena/amd64-supervisor-", 0),
    ("balena/amd64--supervisor", 0),
    ("amd64--supervisor", 0),
    (" amd64-supervisor", 0),
    ("amd64-supervisor ", 0),
    ("amd64 supervisor", 0),
    ("aarch64-supervisor", 1),
    (" balena/amd64-supervisor ", 0),
    ("/amd64-supervisor", 0)
]

for t in matches:
    if t[1] == 1 and re.match(SUPERVISOR_REPOSITORY_RE, t[0]):
        continue
    elif t[1] == 0 and re.match(SUPERVISOR_REPOSITORY_RE, t[0]) is None:
        continue
    else:
        print("Failed on %s match" % (t[0]))
        sys.exit(1)

```

alexgg:
@balena-ci rebase

pdcastro (approved):

PR #247, opened 2021-06-22
Delay populating global partitions cache until methods call

klutchell:
In case the sfdisk command fails we need it to be in a method
that can handle failures, not the container startup cmd.

Connected-to: https://github.com/balena-io/balena-cli/issues/2045
Change-type: patch

### DIFF
--- a/src/preload.py
+++ b/src/preload.py
@@ -886,6 +886,7 @@ def get_docker_storage_driver(docker_service_file_contents):
 
 
 def main_preload(app_data, additional_bytes, splash_image_path):
+    init()
     additional_bytes = round_to_sector_size(ceil(additional_bytes))
     flasher_root = get_partition("flash-rootA")
     if flasher_root:
@@ -913,6 +914,7 @@ def main_preload(app_data, additional_bytes, splash_image_path):
 
 
 def get_image_info():
+    init()
     images, supervisor_version, balena_os_version = (
         get_images_and_supervisor_version()
     )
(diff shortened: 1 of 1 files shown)


src/preload.py
```diff
@@ -913,6 +913,7 @@ def main_preload(app_data, additional_bytes, splash_image_path):
 
 
 def get_image_info():
+    PARTITIONS_CACHE[None] = prepare_global_partitions()
```
pdcastro, line 916: My unease here, if I am reading it right, is that `get_image_info()` and `main_preload()` are two top-level "exported-so-to-speak" methods:
```
methods = {
    "get_image_info": get_image_info,
    "preload": main_preload,
}

if __name__ == "__main__":
    ...
    for line in sys.stdin:
        try:
            data = json.loads(line)
            method = methods[data["command"]]
            result = method(**data.get("parameters", {}))
```

Both methods eventually call `get_partition()`, but only `get_image_info()` initializes `PARTITIONS_CACHE[None]` with this PR's proposed changes. If `main_preload()` was called before or instead of `get_image_info()`, `PARTITIONS_CACHE[None]` would go uninitialized.

For want of a better idea, I could suggest moving `PARTITIONS_CACHE[None] = prepare_global_partitions()` to some top-level `init()` method that both `get_image_info()` and `main_preload()` would call, subject to some "is_initialized" flag to avoid calling `init()` more than once.

pdcastro (changes requested):

pdcastro (approved):

PR #252, opened 2021-07-22
Use a prebuilt image

Page-:


.resinci.yml
```diff
@@ -1,6 +1,11 @@
 ---
 docker:
-  publish: false
+  builds:
+    - path: .
+      dockerfile: Dockerfile
+      docker_repo: balena/balena-preload
+      publish: true
```
klutchell, line 7: This will limit preload to x86_64 only, when today it works on aarch64 and up until recently armv7. Do we have an example of using balenaCI to build multiarch images?

klutchell: Requires https://github.com/product-os/scripts/issues/155

PR #253, opened 2021-07-22
Prefer `this.appId` to `this.application.id`

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #254, opened 2021-07-22
Switch to tmp-promise instead of tmp

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #256, opened 2021-07-22
Make use of async/await to simplify the code

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #257, opened 2021-07-22
Catch errors during build of preload image

klutchell:
Try to catch errors such as https://github.com/balena-io/balena-cli/issues/2290

Change-type: patch

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -276,8 +276,11 @@ class Preloader extends EventEmitter {
 		await new Bluebird((resolve, reject) => {
 			this.docker.modem.followProgress(
 				build,
-				(error) => {
+				(error, output) => {
 					// onFinished
+					if (!error && output) {
+						error = output.pop().error;
+					}
 					if (error) {
 						reject(error);
 					} else {


pdcastro (approved):

lib/preload.js
```diff
@@ -276,8 +276,11 @@ class Preloader extends EventEmitter {
 		await new Bluebird((resolve, reject) => {
 			this.docker.modem.followProgress(
 				build,
-				(error) => {
+				(error, output) => {
 					// onFinished
+					if (!error && output) {
+						error = output.pop().error;
```
pdcastro, line 282: Oh, I have already approved the PR and it is merged, but if it is possible for `output` to be an empty array, this could produce a TypeError:
```
$ node
> let output = []
> if (output) { output.pop().error }
Uncaught TypeError: Cannot read property 'error' of undefined
```
A fix, if needed, might be to test `if (!error && output && output.length)`.

klutchell: It's possible, I will open a new PR for the fixup!

PR #258, opened 2021-07-23
Use class syntax for declaring instance variables

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #261, opened 2021-07-23
Use dd and blkid to get the partition label at an offset into the image

Page-:
This shaves off around 7 seconds for fetchign partition info in my
local testing

Change-type: patch

### DIFF
--- a/src/preload.py
+++ b/src/preload.py
@@ -12,12 +12,12 @@ from math import ceil, floor
 from re import match, search
 from retry.api import retry_call
 from sh import (
+    blkid,
     btrfs,
     dd,
     df,
     docker as _docker,
     dockerd,
-    file,
     fsck,
     losetup,
     mount,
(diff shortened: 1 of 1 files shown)


jakogut:
Does this work for MBR as well as GPT formatted disks?

Page-:
Do you know which version of balenaOS uses MBR/GPT so they can be tested @jakogut ?

jakogut:
@Page- The only gpt formatted images I'm aware of currently are the ones in [balena-generic](https://github.com/balena-os/balena-generic). @alexgg Do you know of any others?

jakogut:
@Page- Ah, the Jetson images are GPT formatted

```
/home/joseph/balena-os/balena-jetson/layers/meta-balena-jetson/recipes-core/images/balena-image.inc
117:PARTITION_TABLE_TYPE = "gpt"
```

PR #268, opened 2022-01-27
Stop relying on the /device-types/v1 endpoints

thgreasi:
Also bumps balena-sdk to v16 (which bumps the node dependency) and TypeScript to v4.5.

Resolves: #269
Change-type: major
See: https://jel.ly.fish/improvement-stop-relying-device-types-v1-device-type-json-unrelated-things-4fbac3c
See: https://jel.ly.fish/brainstorm-topic-cc53b047-fbee-4778-b664-fa7322e9abf4
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

### DIFF
--- a/lib/preload.ts
+++ b/lib/preload.ts
@@ -360,7 +360,14 @@ export class Preloader extends EventEmitter {
 	}
 
 	async _fetchDeviceTypes() {
-		this.deviceTypes = await this.balena.models.config.getDeviceTypes();
+		this.deviceTypes = await this.balena.models.deviceType.getAll({
+			$select: 'slug',
+			$expand: {
+				is_of__cpu_architecture: {
+					$select: 'slug',
+				},
+			},
+		});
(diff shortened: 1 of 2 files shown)


lib/preload.ts
```diff
@@ -826,12 +846,13 @@ export class Preloader extends EventEmitter {
...
-						should_be_running__release: { __id },
-					} = await this.balena.models.application.get(this.appId, {
-						$select: 'should_be_running__release',
-					});
-					releaseFilter.id = __id;
+					const { should_be_running__release } =
+						await this.balena.models.application.get(this.appId, {
+							$select: 'should_be_running__release',
+						});
+					releaseFilter.id = (
```
thgreasi, line 853: This surfaces that we might need to do a follow-up to add an error if the app is set to track the null release.
I guess though that the CLI might already be checking this...

PR #273, opened 2022-05-20
Add support for preloading v3 target state format

pipex:
Supervisor v13 and above uses the newer v3 target state endpoint. This updates the preloading mechanism so `apps.json` is created using this new format for OS images including this newer supervisor.

Change-type: minor

### DIFF
--- a/lib/preload.ts
+++ b/lib/preload.ts
@@ -523,6 +523,17 @@ export class Preloader extends EventEmitter {
 		this.emit('progress', { name, percentage });
 	}
 
+	// Return the version of the target state endpoint used by the supervisor
+	_getStateVersion() {
+		if (this._supervisorLT7()) {
+			return 1;
+		} else if (this._supervisorLT13()) {
+			return 2;
+		} else {
+			return 3;
+		}
+	}
+
 	async _getState() {
 		const uuid = this.balena.models.device.generateUniqueKey();
 		const deviceInfo = await this.balena.models.device.register(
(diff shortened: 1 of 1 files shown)


lib/preload.ts
```diff
@@ -587,16 +609,32 @@ export class Preloader extends EventEmitter {
 					ci.image[0].is_stored_at__image_location.toLowerCase(),
 			});
 		});
+
 		// App from the state endpoint (v1 or v2 depending on the supervisor version).
-		const app = _.values(this.state.local.apps)[0];
 		if (this._supervisorLT7()) {
```
klutchell, line 614:
```suggestion
		if (version === 1) {
```

lib/preload.ts
```diff
@@ -587,16 +609,32 @@ export class Preloader extends EventEmitter {
...
+			const app = _.values(this.state.local.apps)[0];
 			// Pre-multicontainer: there is only one image: use the only image from the state endpoint.
 			images[0].is_stored_at__image_location = app.image.toLowerCase();
 		} else {
+			const _getAppServices = (apps) => {
+				if (this._supervisorLT13()) {
```
klutchell, line 620:
```suggestion
				if (this._getStateVersion() === 2) {
```

lib/preload.ts
```diff
@@ -587,16 +609,32 @@ export class Preloader extends EventEmitter {
 					ci.image[0].is_stored_at__image_location.toLowerCase(),
 			});
 		});
+
 		// App from the state endpoint (v1 or v2 depending on the supervisor version).
-		const app = _.values(this.state.local.apps)[0];
 		if (this._supervisorLT7()) {
+			const app = _.values(this.state.local.apps)[0];
```
klutchell, line 615: Should we adjust `_getAppServices` to support all versions and use it everywhere for consistency?

pipex: Yeah, that would simplify the code. I'll make the change

klutchell:
LGTM. Have you been able to test it locally with a build of the CLI?

You can manually copy these files to the `node_modules` of a balena-cli clone and build it for testing. Or you can open a draft PR of the CLI and pin to this pre-release of the balena-preload package (`balena-preload@12.1.0-preload-v3-02d2264acd2e5ca01bb907d24426c4a3c919839d`).

pipex:
@klutchell I have tested it (using [npm-link](https://docs.npmjs.com/cli/v8/commands/npm-link)). I tested with a v2 supervisor and a v3 supervisor. Both preload apps.json correctly, although the latest supervisor had a couple of preload bugs, hence https://github.com/balena-os/balena-supervisor/pull/1944

klutchell (approved):

PR #277, opened 2022-11-17
fix stdin with docker 20.10.17

aethernet:
Fix regression with docker 20.10.17


aethernet:
Thanks to [PhaseOne](https://github.com/phaseOne) for the PR

Fix https://github.com/balena-io-modules/balena-preload/pull/275

aethernet:
@balena-ci I self-certify!

PR #278, opened 2022-12-16
minor: add .etch preload functions

aethernet:
Prepare .etch preloading


klutchell:
Why are we using the same repo? I would be in favour of a new one since the CLI depends on this and we don't want to include any of the new stuff.

aethernet:
@klutchell at some point the CLI will need to be able to do this

cli already depends on this and etcher-sdk, so it will have the two parts of the process. 
We'll eventually be able to replace the cli preloader with a (transparent) .etch mechanism and completely remove the dependency on a running docker engine

PR #279, opened 2022-12-19
minor: build for every major node version >=12

aethernet:
As node 12 is deprecated and in order to build the CLI on node 14+


aethernet:
@balena-ci I self-certify!

PR #294, opened 2024-07-05
Replace request library in favor of native fetch api

otaviojacobi:
Change-type: patch

### DIFF
--- a/lib/preload.ts
+++ b/lib/preload.ts
@@ -10,7 +10,6 @@ import { promises as fs, constants } from 'fs';
 import * as getPort from 'get-port';
 import * as os from 'os';
 import * as compareVersions from 'compare-versions';
-import * as request from 'request-promise';
 import type {
 	Application,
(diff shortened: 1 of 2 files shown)


lib/preload.ts
```diff
@@ -564,25 +563,32 @@ export class Preloader extends EventEmitter {
...
-			followRedirect,
-			encoding,
+
+		const response = await fetch(`https://${registryUrl}${endpoint}`, {
+			method: 'GET',
+			headers: headers,
+			redirect: followRedirect ? 'follow' : 'manual'
 		});
+
+		if (response.status > 400) {
```
myarmolinsky, line 582: ?
```suggestion
		if (response.status >= 400) {
```

otaviojacobi: in fact, this throw should be gone

otaviojacobi:
superseded by #295

PR #295, opened 2024-07-08
Use the sdk instead of `request` for registry requests

thgreasi:
Depends-on: https://github.com/balena-io-modules/balena-request/pull/184
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/balenaCLI.20bumps/near/450776493

### DIFF
--- a/lib/preload.ts
+++ b/lib/preload.ts
@@ -563,8 +563,12 @@ export class Preloader extends EventEmitter {
 	}
 
 	async registryRequest<RF extends 'json' | 'blob'>(
-		registryUrl: string,
-		endpoint: string,
+		url:
+			| {
+					registryUrl: string;
+					layerUrl: string;
+			  }
+			| string,
 		registryToken: string | null,
 		headers: Record<string, string>,
 		responseFormat: RF,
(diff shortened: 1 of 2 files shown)


lib/preload.ts
```diff
@@ -564,53 +563,65 @@ export class Preloader extends EventEmitter {
...
 			url: `https://${registryUrl}${endpoint}`,
-			headers,
+			headers: {
+				...headers,
+				Authorization: `Bearer ${registryToken}`,
+			},
 			json: decodeJson,
-			simple: false,
-			resolveWithFullResponse: true,
+			...(decodeJson && { responseFormat: 'blob' }),
```
otaviojacobi, line 580: @thgreasi shouldn't this be `...(decodeJson && { responseFormat: 'blob' })` ? My understanding was if we have decodeJson we want JSON output, otherwise, we want blob ?

otaviojacobi:
This works :tada:

otaviojacobi:
superceds #294

otaviojacobi (approved):

PR #312, opened 2026-04-03
Move to ESM, drop support for nodejs < 24.11.0 & update TypeScript to 6.0.2

thgreasi:
Change-type: major
See: https://balena.fibery.io/Work/Project/2389

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "license": "Apache-2.0",
   "author": "Balena Ltd (https://balena.io)",
   "main": "build/preload.js",
+  "type": "module",
   "types": "build/preload.d.ts",
   "engines": {
     "node": "^24.11.0"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,9 +5,10 @@
   "license": "Apache-2.0",
   "author": "Balena Ltd (https://balena.io)",
   "main": "build/preload.js",
+  "type": "module",
   "types": "build/preload.d.ts",
   "engines": {
-    "node": "^22.18.0 || >= 24.0.0"
+    "node": "^24.11.0"
```
Page-, line 11: Any reason not to use `>=`?

package.json
```diff
@@ -35,11 +36,11 @@
   "devDependencies": {
     "@balena/lint": "^7.2.6",
```
Page-, line 37: It probably makes sense to update this to a version that support ts6

